PR #2, opened 2022-12-08
Kuutamo fixes

Mic92:


flake.nix
```diff
@@ -0,0 +1,9 @@
+{
+  description = "Server-optimized nixos configuration";
+
+  outputs = { ... }: {
+    nixosModules = {
+      common = ./profiles/default.nix;
```
Mic92, line 6: Should we rename the profile as well?

PR #6, opened 2022-12-21
Add new profile to create self hosted runners using github app

jfroche:
Copy nixos module and adapt it to use short lived github app access token to register github self hosted runners


zimbatm:
Is there anything missing from this PR?

jfroche:
nop it is working, will rebase and merge tomorrow
I ll push it in nixpkgs for review as well

modules/github-runners.nix
```diff
@@ -0,0 +1,74 @@
+{ config
```
zimbatm, line 1: this file could be moved to modules/github-runnders/default.nix

default.nix
```diff
@@ -16,6 +16,9 @@
   mixins-systemd-boot = import ./mixins/systemd-boot.nix;
   mixins-telegraf = import ./mixins/telegraf.nix;
 
+  # Modules
+  github-runner = import ./modules/github-runners.nix;
```
zimbatm, line 20: If the module is getting upstreamed, let's keep it private and import it in the role directly.

zimbatm (commented):
looking good!

zimbatm (approved):

PR #7, opened 2022-12-21
use systemd-boot everywhere

zimbatm:
Moving us closer to trusted boot

### DIFF
--- a/profiles/common.nix
+++ b/profiles/common.nix
@@ -8,7 +8,7 @@
     ./common/zfs.nix
   ];
 
-  # Use systemd during boot as well
+  # Use systemd during boot as well on systems that do not require networking in early-boot
   boot.initrd.systemd.enable = lib.mkDefault (!config.boot.initrd.network.enable);
 
   # Work around for https://github.com/NixOS/nixpkgs/issues/124215
--- a/profiles/common.nix
+++ b/profiles/common.nix
@@ -9,7 +9,7 @@
   ];
 
   # Use systemd during boot as well
-  boot.initrd.systemd.enable = true;
+  boot.initrd.systemd.enable = lib.mkDefault (!config.boot.initrd.network.enable);
 
   # Work around for https://github.com/NixOS/nixpkgs/issues/124215
   documentation.info.enable = false;
--- a/profiles/common.nix
+++ b/profiles/common.nix
@@ -8,6 +8,9 @@
     ./common/zfs.nix
   ];
 
+  # Use systemd during boot as well
+  boot.initrd.systemd.enable = true;
+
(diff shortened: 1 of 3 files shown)


profiles/common.nix
```diff
@@ -8,6 +8,15 @@
     ./common/zfs.nix
   ];
 
+  # Use systemd during boot as well
+  boot.initrd.systemd.enable = true;
```
Mic92, line 12: We cannot do this as long hetzner exists...

profiles/common.nix
```diff
@@ -8,6 +8,15 @@
...
+
+  # Use EFI boot everywhere
+  boot.loader.efi.canTouchEfiVariables = true;
```
Mic92, line 15: This will fill up your EFI flash on every nixos-rebuild and potentially destroy cheap chips. It should be only done during installation.

Mic92:
I would suggest to make a systemd-boot profile for now and make it optional this way. There also other targets that need uboot or even EFI platforms that do not work with systemd-boot (We have one expensive arm server that does not boot with systemd-boot)

profiles/common.nix
zimbatm: Do you remember what the root cause was?

Mic92: Hetzner does not allow you to boot machines with EFI. I think they now do this if you open a ticket but this is not a good out-of-the-box experience. And than we also have many vps provider that do support EFI as well.

zimbatm: How is having systemd in the initrd linked to EFI?

profiles/common.nix
zimbatm: Good to know, thanks

profiles/common.nix
Mic92: Ah. Sorry. I was referring to systemd-boot being enabled. Not the initrd option.
I think the only reason not to use this yet here is if you have `boot.initrd.network.enable` set i.e. if you try to enter your disk encryption password via ssh, simply because this is not supported yet.

profiles/common.nix
```diff
@@ -8,6 +8,9 @@
     ./common/zfs.nix
   ];
 
+  # Use systemd during boot as well
+  boot.initrd.systemd.enable = true;
```
Mic92, line 12:
```suggestion
  boot.initrd.systemd.enable = lib.mkDefault (!config.boot.initrd.network.enable);
```

profiles/common.nix
```diff
@@ -8,6 +8,9 @@
     ./common/zfs.nix
   ];
 
+  # Use systemd during boot as well
```
Mic92, line 11:
```suggestion
  # Use systemd during boot as well on systems that do not require networking during initrd
```

PR #8, opened 2022-12-21
use systemd-networkd everywhere

zimbatm:
Stop using piles of bash scripts?

### DIFF
--- a/profiles/common.nix
+++ b/profiles/common.nix
@@ -15,8 +15,8 @@
   networking.firewall.allowPing = true;
 
   # Use networkd instead of the pile of shell scripts
-  networking.useNetworkd = true;
-  networking.useDHCP = false;
+  networking.useNetworkd = lib.mkDefault true;
+  networking.useDHCP = lib.mkDefault false;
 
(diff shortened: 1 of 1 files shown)


profiles/common.nix
```diff
@@ -14,6 +14,10 @@
   # Allow PMTU / DHCP
   networking.firewall.allowPing = true;
 
+  # Use networkd instead of the pile of shell scripts
+  networking.useNetworkd = true;
+  networking.useDHCP = false;
```
Mic92, line 19: Since this is also a desktop profile, how does this interact with network-manager?

zimbatm: I will try it and report back

zimbatm: Seems to be working fine. NetworkManager is running side-by-side and it still used to connect to the wifi, acquire DHCP, ...

Mic92, line 19: Could be dangerous to overwrite defaults that people are already used to.

```suggestion
  networking.useNetworkd = lib.mkDefault true;
  networking.useDHCP = lib.mkDefault false;
```

PR #9, opened 2022-12-23
improve defaults for serial console

Mic92:


profiles/common/serial.nix
```diff
@@ -0,0 +1,17 @@
+{ lib, pkgs, ... }:
+{
+  # Configure so that we can use serial consoles.
+  # This is for example important for IPMI SOL console redirection or GPIO-based ttl-converter
+  boot.kernelParams =
+    [ "console=ttyS0,115200" ] ++
+    (lib.optional (pkgs.stdenv.isAarch32 || pkgs.stdenv.isAarch64) "console=ttyAMA0,115200") ++
+    (lib.optional (pkgs.stdenv.isRiscv64) "console=ttySIF0,115200") ++
+    [ "console=tty0" ];
```
Mic92, line 9: Nixos seems to use a different order compared to other linux distributions.
I have not fully understood yet how and why order matters but I think to remember that it can break things...

PR #10, opened 2022-12-24
add nginx role

Mic92:


roles/nginx.nix
```diff
@@ -0,0 +1,37 @@
...
+    recommendedOptimisation = lib.mkDefault true;
+    recommendedProxySettings = lib.mkDefault true;
+    recommendedTlsSettings = lib.mkDefault true;
+
+    # Nginx sends all the access logs to /var/log/nginx/access.log by default.
+    # instead of going to the journal!
+    commonHttpConfig = ''
+      access_log syslog:server=unix:/dev/log;
+      # Should we have this?
+      #  add_header Strict-Transport-Security 'max-age=31536000; includeSubDomains; preload' always;
```
zimbatm, line 17: Only if TLS/ACME is enabled

flake.nix
```diff
@@ -8,10 +8,14 @@
...
     nixosModules = {
+      # Profiles
       common = import ./profiles/common.nix;
       server = import ./profiles/server.nix;
-      github-actions-runner = import ./roles/github-actions-runner.nix;
       telegraf = import ./profiles/telegraf.nix;
+
+      # Roles
+      github-actions-runner = import ./roles/github-actions-runner.nix;
+      nginx = import ./roles/nginx.nix;
```
zimbatm, line 18: I think it goes into profiles since it might be used by other modules.

roles/nginx.nix
Mic92: Only issue is that it's hard to tell when this is the case. Servers might have many vhosts and with conflicting settings.

Mic92: Leaving it out know, since enabling hsts can be difficult when you are trying to turn it off again.

PR #12, opened 2022-12-28
add flake profile

Mic92:


profiles/common/flake.nix
```diff
@@ -0,0 +1,66 @@
...
+      description = lib.mdDoc ''
+        Flake that the nixos contains the nixos configuration.
+      '';
+    };
+
+    symlinkFlake = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = lib.mdDoc ''
+        Symlinks the the flake the system was built with to `/run/current-system`
```
zimbatm, line 21:
```suggestion
        Symlinks the flake the system was built with to `/run/current-system`
```

zimbatm (approved):

profiles/common/flake.nix
```diff
@@ -0,0 +1,66 @@
...
+{
+  options.srvos = {
+    flake = lib.mkOption {
+      # FIXME what is the type of a flake?
+      type = lib.types.nullOr lib.types.raw;
+      default = null;
+      description = lib.mdDoc ''
+        Flake that the nixos contains the nixos configuration.
+      '';
+    };
```
zimbatm, line 15: What do you think of relying on `_module.args.inputs` to exist as a convention instead?

Mic92: Is there a way to detect if `_module.args.inputs` exists? otherwise I am unsure how to make this the default without breaking evaluation

zimbatm: with `_module.args ? inputs`, no?

zimbatm: but I agree that the relationship is more implicit so it would have to be documented properly

Mic92: We also need `_module.args ? self`. I can make the option default to set...

Mic92: It does not seem to work. I cannot access _module.args.

PR #17, opened 2022-12-29
Add hardware

zimbatm:
Introduce opinionated hardware configurations for cloud environments.

### DIFF
--- a/hardware/hetzner-cloud/default.nix
+++ b/hardware/hetzner-cloud/default.nix
@@ -1,0 +1,39 @@
+{ config, modulesPath, ... }:
+{
+  imports = [
+    ../../mixins/cloud-init.nix
+    "${modulesPath}/profiles/qemu-guest.nix"
+  ];
+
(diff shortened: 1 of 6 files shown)


hardware/hetzner-cloud/default.nix
```diff
@@ -0,0 +1,29 @@
...
+    boot.cleanTmpDir = true;
+    boot.growPartition = true;
+    boot.loader.grub.device = "/dev/sda";
+
+    fileSystems."/" = { device = "/dev/sda1"; fsType = "ext4"; };
+
+    networking.useNetworkd = true;
+    networking.useDHCP = false;
+
+    systemd.network.networks."10-uplink" = {
```
Mic92, line 18: An assertion to add an ipv6 address would be nice here. Otherwise we have servers with broken aaaa records again.

zimbatm: added an assertion

PR #18, opened 2022-12-30
expose modules list

zimbatm:
Make the modules list available outside of flakes and prefix them.


zimbatm:
I think we'll need some sort of structure like that. What do you think?

default.nix
```diff
@@ -0,0 +1,13 @@
+{
+  # Profiles
+  profiles-common = import ./profiles/common.nix;
+  profiles-server = import ./profiles/server.nix;
+  profiles-desktop = import ./profiles/desktop.nix;
+
+  profiles-efi = import ./profiles/efi.nix;
+  profiles-nginx = import ./profiles/nginx.nix;
+  profiles-telegraf = import ./profiles/telegraf.nix;
```
Mic92, line 9: Mhm. I find the `profiles-` prefix a bit verbose.

default.nix
```diff
@@ -0,0 +1,13 @@
+{
+  # Profiles
+  profiles-common = import ./profiles/common.nix;
+  profiles-server = import ./profiles/server.nix;
+  profiles-desktop = import ./profiles/desktop.nix;
```
Mic92, line 5: I think these are actually our main modules while the below behave more like mixins that you can add on top.

Mic92: Something that indicates that you should add one of these and than some of the below would be nice.

Mic92: On the other hand already the categories profiles and roles on its own are confusing enough..

zimbatm: Good point. I moved things around a bit.

PR #22, opened 2023-01-02
add more terminfo packages but hide them when cross compiling

Mic92:


server.nix
```diff
@@ -12,9 +12,15 @@
...
     pkgs.htop
     pkgs.jq
-    pkgs.termite.terminfo
     pkgs.tmux
     pkgs.vim
+    # avoid compiling desktop stuff when doing cross nixos
+  ] ++ lib.optionals (pkgs.stdenv.hostPlatform == pkgs.stdenv.buildPlatform) [
+    # various terminfo packages
+    pkgs.termite.terminfo
+    pkgs.wezterm.terminfo
```
SuperSandro2000, line 21: FYI At least wezterm's terminfo is cross friendly with minimal dependencies https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/terminal-emulators/wezterm/default.nix#L123-L131 because if you ssh into a machine with wezterm as your terminal programs will be confused about the missing data.

Mic92: okay that one might be fine to move than.

Mic92: https://github.com/numtide/srvos/pull/23

PR #26, opened 2023-01-04
hardware-hetzner-online: add available kernel modules

zimbatm:
And enable redistributables by default.

### DIFF
--- a/hardware/hetzner-online/amd.nix
+++ b/hardware/hetzner-online/amd.nix
@@ -2,6 +2,7 @@
   imports = [ ./. ];
 
   boot.kernelModules = [ "kvm-amd" ];
-  hardware.cpu.amd.updateMicrocode =
-    lib.mkDefault config.hardware.enableRedistributableFirmware;
+  boot.initrd.availableKernelModules = [ "xhci_pci" "ahci" "nvme" "usbhid" ];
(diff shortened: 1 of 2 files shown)


hardware/hetzner-online/amd.nix
```diff
@@ -2,6 +2,7 @@
   imports = [ ./. ];
 
   boot.kernelModules = [ "kvm-amd" ];
-  hardware.cpu.amd.updateMicrocode =
-    lib.mkDefault config.hardware.enableRedistributableFirmware;
+  boot.initrd.availableKernelModules = [ "xhci_pci" "ahci" "nvme" "usbhid" ];
+  hardware.cpu.intel.updateMicrocode = lib.mkDefault config.hardware.enableRedistributableFirmware;
+  hardware.enableRedistributableFirmware = lib.mkDefault true;
```
Mic92, line 7: This is already in here: https://github.com/numtide/srvos/blob/master/hardware/hetzner-online/default.nix#L18

zimbatm: what about `usbhid`?

Mic92: I am pretty sure this is just miss-detection. We don't need usb human interface devices on hetzner. I think this might be also only added in case someone needs a usb keyboard for entering decryption passwords on a desktop machine.

zimbatm: Ok then this PR is not needed

PR #28, opened 2023-01-10
hardware-vultr: add new profile

zimbatm:
I think we need to introduce disk partitioning soon

### DIFF
--- a/hardware/vultr/default.nix
+++ b/hardware/vultr/default.nix
@@ -1,0 +1,34 @@
+{ lib, ... }:
+{
+  imports = [ ../../mixins/cloud-init.nix ];
+
+  boot.kernelParams = [ "vultr" ]; # needed for cloud-init on baremetal server
+
+  services.cloud-init.config = ''
+    system_info:
+      distro: nixos
+      network:
+        renderers: [networkd]
+    users:
+       - root
+
+    disable_root: false
+    preserve_hostname: false
+
+    datasource:
+      Vultr:
+        url: http://169.254.169.254
+        retries: 30
+        timeout: 10
+        wait: 5
+
+    cloud_init_modules:
+     - set_hostname
+     - update_hostname
+     - update_etc_hosts
+    cloud_config_modules: []
+    cloud_final_modules: []
+  '';
+  systemd.services.cloud-init.enable = false;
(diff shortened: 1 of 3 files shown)


hardware/vultr/default.nix
```diff
@@ -0,0 +1,34 @@
+{ lib, ... }:
+{
```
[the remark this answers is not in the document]
zimbatm: It's not in the original code either.

imsofi:
@zimbatm You inspired me to work a bit on cloud-init with this PR. Wanted to bring back some information for how to get this functionality to work correctly without a bunch of retries: https://github.com/NixOS/nixpkgs/issues/215571

zimbatm:
Thanks for the heads-up :)

nixos/hardware/vultr/bare-metal.nix
```diff
@@ -0,0 +1,42 @@
...
+
+      cloud_init_modules:
+       - set_hostname
+       - update_hostname
+       - update_etc_hosts
+      cloud_config_modules: []
+      cloud_final_modules: []
+    '';
+  };
+  systemd.services.cloud-init.enable = false;
```
Mic92, line 40: Can you elaborate with cloud-init.service is disabled here? Does undo the configuration above?

zimbatm: @nitrotm do you remember why it was set that way?

nitrotm: right, that was because the cloud-init local setup was working (and enough to set hostname) but the rest was failing.

nitrotm: something about the Vultr datasource not being able to setup more advanced network parameters if I recall properly; I am not sure if that's still actual.

zimbatm: thanks, I will try to reproduce the issue then

zimbatm:
needs to be rebased once https://github.com/NixOS/nixpkgs/pull/229111 and https://github.com/NixOS/nixpkgs/pull/229103 have been merged

zimbatm:
bors merge

PR #29, opened 2023-01-10
desktop: enable pipewire by default

zimbatm:


desktop/pipewire.nix
```diff
@@ -0,0 +1,10 @@
+# Use pipewire instead of PuleseAudio. See https://nixos.wiki/wiki/PipeWire
+{
+  security.rtkit.enable = true;
+  services.pipewire = {
+    enable = true;
+    alsa.enable = true;
+    alsa.support32Bit = true;
```
Mic92, line 7: Probably does not make sense on other architectures than x86:

```suggestion
    alsa.support32Bit = pkgs.stdenv.isx86_64;
```

desktop/pipewire.nix
```diff
@@ -0,0 +1,10 @@
+# Use pipewire instead of PuleseAudio. See https://nixos.wiki/wiki/PipeWire
+{
+  security.rtkit.enable = true;
```
Mic92, line 3:
```suggestion
  security.rtkit.enable = lib.mkDefault true;
```

desktop/pipewire.nix
```diff
@@ -0,0 +1,10 @@
...
+  services.pipewire = {
+    enable = true;
```
Mic92, line 5:
```suggestion
    enable = lib.mkDefault true;
```

desktop/pipewire.nix
```diff
@@ -0,0 +1,10 @@
...
+    alsa.enable = true;
```
Mic92, line 6:
```suggestion
    alsa.enable = lib.mkDefault true;
```

desktop/pipewire.nix
```diff
@@ -0,0 +1,10 @@
...
+    alsa.support32Bit = true;
+    pulse.enable = true;
```
Mic92, line 8:
```suggestion
    pulse.enable = lib.mkDefault true;
```

desktop/pipewire.nix
Mic92: But also, what applications are still 32-bit? I have not any in a while.

zimbatm: Good call. I added this for Steam and could be enabled separately.

Mic92: I see. I guess with games we will have 32-bit at least for another decade.

zimbatm:
bors merge

Mic92:
We need to whitelist bors...

zimbatm:
bors merge

PR #32, opened 2023-01-15
well-known-hosts: define hostNames instead of extraHostNames

xddxdd:
This gets rid of the dummy hostnames of `github-rsa`, `github-ed25519`, etc from the `/etc/ssh/ssh_known_hosts` file.


Mic92:
bors merge

PR #38, opened 2023-01-20
add trusted caches mixins

Mic92:
This adds a list of common trust-worthy caches. It's not enabled by default since it's debatable what caches should be enabled by default.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -103,6 +103,14 @@
             self.nixosModules.mixins-terminfo
           ];
         };
+        example-mixins-trusted-nix-caches = nixpkgs.lib.nixosSystem {
+          system = "x86_64-linux";
+          modules = [
+            dummy
+            fake-hardware
+            self.nixosModules.mixins-trusted-nix-caches
+          ];
+        };
         example-mixins-nginx = nixpkgs.lib.nixosSystem {
           system = "x86_64-linux";
           modules = [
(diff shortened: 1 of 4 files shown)


mixins/trusted-caches.nix
```diff
@@ -0,0 +1,9 @@
+{
+  # Caches in trusted-substituters can be used by unprivileged users i.e. in
+  # flakes but are not enabled by default.
+  nix.settings.trusted-substituters = [
+    "https://cache.garnix.io"
+    "https://numtide.cachix.org"
+    "https://nix-community.cachix.org"
+  ];
```
zimbatm, line 8: How about adding the public keys as well?

flake.nix
```diff
@@ -103,6 +103,14 @@
             self.nixosModules.mixins-terminfo
           ];
         };
+        example-mixins-trusted-caches = nixpkgs.lib.nixosSystem {
+          system = "x86_64-linux";
+          modules = [
+            dummy
+            fake-hardware
+            self.nixosModules.mixins-trusted-caches
```
zimbatm, line 111:
```suggestion
            self.nixosModules.mixins-trusted-nix-caches
```
To be a bit more descriptive

zimbatm (commented):
Don't forget to amend the README as well

zimbatm (approved):
bors merge

PR #56, opened 2023-02-03
add nix-remote-builder role

zimbatm:


DavHau:
We deployed our first builder successfully with this. Though I'm not sure about the garbage collection settings. We still seem to have issues regarding invalid paths.
Anyways let's merge this and fix it up later

nixos/roles/nix-remote-builder.nix
```diff
@@ -0,0 +1,36 @@
...
+    # Randomize GC to avoid thundering herd effects.
+    systemd.timers.nix-gc.timerConfig.RandomizedDelaySec = lib.mkForce "1800";
+
+    # Allow more open files for non-root users to run NixOS VM tests.
+    security.pam.loginLimits = [
+      { domain = "*"; item = "nofile"; type = "-"; value = "20480"; }
+    ];
+
+    # Give restricted SSH access to the build scheduler
+    users.extraUsers.root.openssh.authorizedKeys.keys = map
```
DavHau, line 30: We should probably not use root to do remote builds

DavHau (approved):

nixos/roles/nix-remote-builder.nix
zimbatm: Probably. I think there was some performance reason to use root, so it bypasses the nix-daemon entirely. I know that remote building requires trusted users at a minimum.

zimbatm:
bors merge

steveeJ:
the builder deployed with this module produces invalid outputs, see [this log](https://github.com/holochain/holochain/actions/runs/4091509950/jobs/7055645820#step:6:1197). the adjacent jobs run the same derivation just on a darwin remote builder set up differently. from my naive understanding the obvious difference is that the darwin remote builder uses ssh-ng and this builder uses ssh as a protocol.

nixos/roles/nix-remote-builder.nix
```diff
@@ -0,0 +1,36 @@
...
+      description = "SSH public keys of the central build scheduler";
+      type = lib.types.listOf lib.types.str;
+    };
+  };
+
+  config = {
+    # Garbage-collect often
+    nix.gc.automatic = true;
+    nix.gc.dates = "*:45";
+    nix.gc.options = ''--max-freed "$((128 * 1024**3 - 1024 * $(df -P -k /nix/store | tail -n 1 | ${pkgs.gawk}/bin/awk '{ print $4 }')))"'';
```
SuperSandro2000, line 19: @zimbatm what did you try to calculate here? When I run the command on one of my machines the result is negative.

DavHau: This command ensures that after the GC ran, 128 GB of disk space are free. If your free disk space is already more than 128G, then the calculated number will be nagative, in which case the GC won't delete any paths. This is expected.

PR #57, opened 2023-02-04
fix: make experimental-features a list

DavHau:
This fixes the problem, that you cannot add to more experimental features in another config


zimbatm:
bors merge

PR #67, opened 2023-02-13
docs: explain the _file attribute

### LINKED ISSUE #61: Document use of `_file`
**Is your feature request related to a problem? Please describe.**

In commit 43d2842c16c897c901d6f004cdb63c51092e55b4 all nix files were chosen to be self-referenced using `_file` attribute to be better supported by `nix flake check`.

**Describe the solution you'd like**

I think it would be important to document why this choice was made for newcomers, for example a note of its use in `README.md`, and importantly, any possible open issues that track this issue upstream.

**Describe alternatives you've considered**

Not documenting its usage, risking teams not applying this pattern correctly, or even wanting it removed because they do not see its value for being there.

**Additional context**

When reading over the source code, especially if you are new to nix, the use of `_file = file.nix` may seem very counter intuitive and strange to read. Since surely nix knows where its own files are, right?

zimbatm:
Fixes #61

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -1,9 +1,9 @@
 # Table of Content
 
 - [Hello](./hello.md)
-- [Design](./design.md)
 - [Modules](./modules.md)
     - [General](./modules/general.md)
     - [Hardware](./modules/hardware.md)
     - [Mixins](./modules/mixins.md)
     - [Roles](./modules/roles.md)
+- [Design](./design.md)
--- a/docs/design.md
+++ b/docs/design.md
@@ -2,3 +2,9 @@
 
 This page explains some of the design decisions that have been made trough
 this project.
+
+## `_file` attribute in the NixOS modules
+
+You might see that all the NixOS modules in this repo have a `_file` attribute, which is quite uncommon to see. The reason we do this is to improve the error reporting whenever a type issue happens.
+
+Normally that attribute is set by the recursive module `imports` traversal, when `imports` is a list of paths. The issue is that `nix flake check` expects all the modules attached to the `nixosModules` attributes to be functions and not paths. This forces us to use the `import` keyword, and that loses the reference to the path. So we add them back manually.


Mic92:
bors merge

zimbatm:
I wanted to wait on imsofi's feedback before merging. But I can send another PR.

imsofi:
It just hit me that [default.nix](https://github.com/numtide/srvos/blob/f9c21e31ef5a80aaa6b606d9a91e94e0fbc548d5/default.nix#L3) uses `import ./file.nix` instead of `imports = [ ./file.nix ]`.

I was long lost where exactly the `import` keyword was coming from in the explanation, and thought it was meant as an internal part of how `nix flake check`'s worked. As it was what got pointed to by name, rather than the `default.nix` file.

Some ideas would likely be to explain that `imports` is [a handy feature of NixOS's module feature](https://discourse.nixos.org/t/import-list-in-configuration-nix-vs-import-function/11372/3) which sets up a human facing interface around the import function. While `import` essentially works as "move the described file's content to the location from where its imported at".

Maybe emphasise further with an example to `nix flake check` expecting a full module as `nixosModules.name = { config }: { options = {}; config = {}; };`, as `default.nix` is still a bit vague. Should probably have a note it is only needed for the first import of the file, and not all files should include `_file`, as you can start using the `imports` feature after the top-level module is made.

Hope that helps!

imsofi:
This seems to hold a lot more on the inner workings of imports, and is quite the nuanced topic relating to the difference between `imports` and `import`. Where the use of `import` can be seen as "improper" by Nix, leading to a situation of reading source code from a location it does not know. All of which propagates upwards to the `nix flake check` command.

Think a major help in readability for newcomers would be to focus on that difference, give a simple example, show how it shows up as an innocent mistake, which finally leads to mysterious error messages without a file in `nix flake check`.

This recommendation is definitely more of a article/book format, rather than documentation as to why a choice was made. So do feel free to take your own approach with what you feel would be the best fit for the document.

zimbatm:
Thanks @imsofi, this is great feedback! I will expand the docs with it.

zimbatm:
@imsofi is that better? https://github.com/numtide/srvos/commit/7db80eca6c7a32ca38d235b22674a53f24d68567

imsofi:
Besides some spelling, and the confusing `(2)` and `(1)` referencing to `imports` and `import`. Overall looks great!

PR #69, opened 2023-02-16
proposal: predictable module locations

DavHau:
There is a 1 to 1 mapping of files located at:
`modules/<type>/<name>[.nix]`
to output attributes like:
`modules.<type>.<name>`

flake-parts is used to wire things together

The the `all-modules.nix` which auto-exports all modules is copied from github:DavHau/flake-templates#falke-parts

Shortcomings:
This design doesn't cover how internal, shared modules should be managed. An example would be the previous `hardware-hetzner-online` which is the base for `hardware-hetzner-online-intel` and `hardware-hetzner-online-amd`. If this internal module would be put in the top of `modules/nixos/`, it would be exported automatically, which is not desired. As of now I moved it inside `modules/nix/hardware-hetzner-online-intel/base.nix`.

Compatibility:
 - flake-compat ensures that the default.nix and flake.nix are in sync and have the same outputs.
 - the schema is future compatible with https://github.com/NixOS/nix/issues/6257
 - outputs for the current schema `nixosModules.<name>` etc. are also contained


modules/flake-parts/_template.nix
```diff
@@ -0,0 +1,8 @@
+{ self, lib, ... }: {
+  perSystem = { config, self', inputs', pkgs, ... }: {
+
+  };
+  flake = {
+
+  };
+}
```
zimbatm, line 8: I think this file is unused

flake.nix
```diff
@@ -3,180 +3,25 @@
...
-                  "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOuiDoBOxgyer8vGcfAIbE6TC4n4jo8lhG9l01iJ0bZz zimbatm@no1"
-                ];
-              }
-            ];
-          };
-        };
+      imports = [
+        ./modules/flake-parts/all-modules.nix
+        ./modules/flake-parts/docs.nix
+      ];
```
zimbatm, line 25: What if all-modules.nix was also tasked with finding packages in the build tree and mapping them back to packages with a 1:1 mapping?

Then it would look like a flake version of [bld](https://github.com/numtide/bld/blob/main/lib/buildTree.nix)

flake.nix
```diff
@@ -3,180 +3,25 @@
...
-              {
-                roles.nix-remote-builder.schedulerPublicKeys = [
-                  "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIOuiDoBOxgyer8vGcfAIbE6TC4n4jo8lhG9l01iJ0bZz zimbatm@no1"
-                ];
-              }
-            ];
-          };
-        };
+      imports = [
+        ./modules/flake-parts/all-modules.nix
```
zimbatm, line 23: Let's move this out of the ./modules folder to make the DAG non-cyclic.

modules/flake-parts/nixosConfigurations.nix
```diff
@@ -0,0 +1,177 @@
...
+  {
+    # General
+    example-common = nixosSystem {
+      system = "x86_64-linux";
+      modules = [
+        dummy
+        fake-hardware
+        self.nixosModules.common
+      ];
+    };
```
zimbatm, line 43: It would be cool if those could live near their respective modules.

zimbatm (commented):
I think you're onto something. Some adapter that maps a folder structure to flake attributes.

flake.lock
```diff
@@ -16,8 +49,28 @@
         "type": "github"
       }
     },
+    "nixpkgs-lib": {
```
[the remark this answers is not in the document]
DavHau: Following doesn't work because flake-parts is using `?dir=lib`

modules/flake-parts/_template.nix
DavHau: That is intentional. In case someone wants to create a new flake-parts module here, they just need to copy the _template.nix.
Do you still prefer removing it?

flake.nix
DavHau: Oh it already does automatically import it. This line ended up there by accident. I will remove it.

flake.lock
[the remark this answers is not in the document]
zimbatm: Ok, too bad ...

flake.nix
zimbatm: Yeah, it's more for the mental model, so the user doesn't have to dig into the code

modules/flake-parts/_template.nix
zimbatm: I don't have a preference right now. Is that a convention that exists elsewhere?

flake.nix
DavHau: Sorry, I skimmed your message a bit too quick.

So far it assumes that everything in the tree is a module of some kind.
As long as all packages are declared as flake parts modules under `./modules/flake-parts`, then they will be automatically imported.

We could extend the logic to discover package function files as well. We would either have to come up with a heuristic how to distinguish package functions from modules, or define where package functions can be located.

Personally I prefer to just declare all packages inside flake parts modules or as individual drv-parts modules.

modules/flake-parts/_template.nix
DavHau: No, it's just a personal habit with the goal of increasing my productivity when creating new modules.

modules/flake-parts/nixosConfigurations.nix
DavHau: We could make each module a directory and put an `example.nix` inside it.

flake.nix
```diff
@@ -3,180 +3,24 @@
 
   inputs = {
     nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable-small";
+    flake-compat = {
+      url = "github:edolstra/flake-compat";
+      flake = false;
+    };
```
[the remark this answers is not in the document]
Mic92: Ah, this is to make evaluation pure...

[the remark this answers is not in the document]
DavHau: We can remove the flake-compat input and fetch it via builtins. Having it as a flake input just makes it easier to keep track of the dependency and update it.

zimbatm:
@DavHau I pushed a second proposal that solves the shortcoming of the initial one, WRT file layout. I also moved files back a bit to avoid the deep layout

zimbatm:
bors merge

PR #72, opened 2023-02-20
feat(github-runners): add option to install default packages

jfroche:
Some workflows assume packages to be installed by default (e.g. bld). Sometime helpful to not run devshell in all workflows.

### DIFF
--- a/nixos/roles/github-actions-runner.nix
+++ b/nixos/roles/github-actions-runner.nix
@@ -86,6 +86,15 @@ in
         type = lib.types.str;
       };
     };
+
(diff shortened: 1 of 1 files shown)


zimbatm (approved):
👍

nixos/roles/github-actions-runner.nix
```diff
@@ -111,7 +120,7 @@ in
             pkgs.nix
```
Mic92, line 120:
```suggestion
            config.nix.package
```

Mic92:
bors merge

Issue #79, opened 2023-02-27
Drop all flake inputs


Mic92:
We currently have the following 3 dependencies:

- nixpkgs (only used for testing)
- flake-parts (Used but not really needed since we don't have to write the normal boiler code for to access `packages.<system>` for nixos modules)
- flake-compat (not used in the flake but by the old model)

I think we can move `nixpkgs` and `flake-compat` to a subflake only used for development. In this new flake we can also add things like linters and auto formatters needed by the project without bloating up the dependency trees

Advantages of this move: Usually you want to set up `srvos.follow.<input>` for many dependencies in your project. Not having those makes it easier to use srvos. There is a point to make that some users might want to use our tested nixpkgs version, however in practice I see it more often that users need to update nixpkgs with their own pace (or maybe their own fork) rather relying on a third party to make a move.

Mic92:
fixed in https://github.com/numtide/srvos/pull/80

PR #80, opened 2023-02-27
No parts

zimbatm:
Fixes 2/3 of #79

Let's keep nixpkgs as having a blessed version of nixpkgs is still useful in my opinion. There is also the fact that working with sub-flakes is also too cumbersome right now.


Mic92:
bors merge

Mic92:
Thanks.

PR #82, opened 2023-02-27
flake: add bors TOML check

zimbatm:
A new trick to keep the bors.toml in sync between Garnix and the flake content.


zimbatm:
bors merge

PR #85, opened 2023-03-04
github-actions-runner: drop kvm extra group

Mic92:
the systemd option is invalid and the device should be world-readable.


zimbatm:
@jfroche do you remember the purpose of giving access to `/dev/kvm`? It's not needed to run the NixOS tests given that they are handled by the /nix daemon.

jfroche:
I need that option to be able to run nixosTest driver by the user to get the result output like this:
```
${nixosTest.driver}/bin/nixos-test-driver --no-interactive -o result
```
usually to create a nice test report with https://github.com/dorny/test-reporter

jfroche:
bors merge

PR #87, opened 2023-03-07
fix(github-runner): unregister existing offline runner using GitHub App token

jfroche:
The GitHub self-hosted runner service can unregister existing GitHub runners if their token still exists in the runner's state directory. However, if the runner is still registered and the token no longer exists, the service is unable to unregister the runner, and it cannot be started again as it is already registered.

Rather than relying on the token stored in the runner's state directory, this change enables the service to unregister an existing offline runner using the GitHub App token.


Mic92:
bors merge

PR #97, opened 2023-03-18
Don't use users.extraUsers alias

SuperSandro2000:
from the description of the option:
  Alias of users.users.


Mic92:
bors merge

PR #100, opened 2023-03-24
well-known-hosts: update rsa pubkey for GitHub

xddxdd:
GitHub changed their RSA pubkey on March 23. <https://github.blog/2023-03-23-we-updated-our-rsa-ssh-host-key/>


zimbatm (approved):
thanks!

bors merge

PR #105, opened 2023-03-28
feat: add nix formatter

jfroche:


flake.nix
```diff
@@ -1,9 +1,15 @@
 {
   description = "Server-optimized NixOS configuration";
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable-small";
+    treefmt-nix = {
+      url = "github:numtide/treefmt-nix";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
```
zimbatm, line 8: This means that all the users of srvos now have an additional flake input in their lockfile. Is it possible to pull it differently?

zimbatm: I tried to work around this issue in #109

PR #107, opened 2023-03-31
feat(github-runner): add role option to add extra labels

jfroche:
Introduces the `extraLabels` option to the GitHub action runners role, enabling users to target specific runners using custom labels.

### DIFF
--- a/nixos/roles/github-actions-runner.nix
+++ b/nixos/roles/github-actions-runner.nix
@@ -97,6 +97,15 @@ in
       default = [ ];
     };
 
+    extraLabels = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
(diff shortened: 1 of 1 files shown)


nixos/roles/github-actions-runner.nix
```diff
@@ -97,6 +97,15 @@ in
       default = [ ];
     };
 
+    extraLabels = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      description = lib.mdDoc ''
+        Extra labels to add to the runners to be able to target them.
+      '';
+      default = [ ];
```
Mic92, line 105: Maybe make the default easy to override?

```suggestion
      default = ["nix"];
```

PR #119, opened 2023-04-11
feat(github-runner): add extra read write paths option

jfroche:
GitHub Runner services are configured to makes the entire system read-only by default. Add a github-action-runner role option to allow specific paths to remain writable for the services.

To explicitely share files between runners in these paths, a shared group ID is also be added to the runners.

### DIFF
--- a/nixos/roles/github-actions-runner.nix
+++ b/nixos/roles/github-actions-runner.nix
@@ -81,6 +81,19 @@ in
       type = lib.types.int;
     };
 
+    extraReadWritePaths = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      description = lib.mdDoc ''
+        The GitHub Runner service is configured to makes the entire system read-only by default.
+        This option can be used to allow specific paths to remain writable for the service.
+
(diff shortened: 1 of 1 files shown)


nixos/roles/github-actions-runner.nix
```diff
@@ -81,6 +81,18 @@ in
...
     };
 
+    extraReadWritePaths = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      description = lib.mdDoc ''
+        The GitHub Runner service is configured to makes the entire system read-only by default.
+        This option can be used to allow specific paths to remain writable for the service.
+
+        Warning: As different unit might get the same UID/GID assigned later on, the files created in those paths are
+        eventually accessible to all github runners.
```
Mic92, line 91:
```suggestion
        Warning: As different unit might get the same UID/GID assigned later on, the files created in those paths are 
        eventually accessible to all github runners. 
        Do not use this option if different github action pipelines should not be able to access state between each other (for security reason)
```

jfroche:
bors merge

PR #127, opened 2023-04-17
server: disable documentation by default

zimbatm:
Keep the evaluation and closure size small

### DIFF
--- a/nixos/server/default.nix
+++ b/nixos/server/default.nix
@@ -17,6 +17,7 @@
     pkgs.vim
   ];
 
+  # Notice this also disables --help for some commands such es nixos-rebuild
   documentation.enable = lib.mkDefault false;
 
   programs.vim.defaultEditor = true;


nixos/server/default.nix
```diff
@@ -17,6 +17,8 @@
     pkgs.vim
   ];
 
+  documentation.enable = lib.mkDefault false;
```
Mic92, line 20:
```suggestion
  # Notice this also disables --help for some commands such es nixos-rebuild
  documentation.enable = lib.mkDefault false;
```

Mic92: I had the idea to turn this into a standalone package:

```
$ nix repl
Welcome to Nix 2.13.3. Type :? for help.

nix-repl> :lf .#                                                         
Added 19 variables.

nix-repl> :b nixosConfigurations.eve.config.system.build.manual.manpages

This derivation produced the following outputs:
  out -> /nix/store/919zbm49lcizzdd66myyrgwzkb8fm444-nixos-manpages

nix-repl> 
```

similar to nix-index-database.

zimbatm:
bors merge

nixos/server/default.nix
Mic92: I started some repo, will update you once I got it running.

PR #130, opened 2023-04-22
server: disable all documentation, disable fontconfig

zowoq:


nixos/server/default.nix
```diff
@@ -23,6 +23,9 @@
   documentation.man.enable = lib.mkDefault false;
   documentation.nixos.enable = lib.mkDefault false;
 
+  # No need for fonts on a server
+  fonts.fontconfig.enable = false;
```
zowoq, line 27: https://github.com/NixOS/nixpkgs/blob/e025c6f560d19c5c72bfcc770bb5e9d18295d0d4/nixos/modules/config/fonts/fontconfig.nix#L267

> If enabled, a Fontconfig configuration file will be built
pointing to a set of default fonts.  If you don't care about
running X11 applications or any other program that uses
Fontconfig, you can turn this option off and prevent a
dependency on all those fonts.

otavio (approved):

zimbatm (approved):
cheers!

bors merge

PR #147, opened 2023-05-12
telegraf: make mdstat monitoring option

Mic92:
It makes telegraf otherwise errors if /proc/mdstat is not available.


Mic92:
cc @zowoq (we might need to explicitly enable this in infra)

Mic92:
bors merge

PR #173, opened 2023-06-26
common: default to boot.tmp.cleanOnBoot=true

phaer:
Hello! Just noticed that `boot.tmp.cleanOnBoot` is only enabled on hetzner-cloud machines in srvos. Is there any drawback in doing so on all machines by default or shall we change that? :)

### DIFF
--- a/nixos/common/default.nix
+++ b/nixos/common/default.nix
@@ -35,4 +35,7 @@
 
   # Allow sudo from the @wheel group
   security.sudo.enable = true;
+
+  # Ensure a clean & sparkling /tmp on fresh boots.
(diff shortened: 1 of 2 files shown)


phaer:
CI is supposedly broken because of https://github.com/numtide/mkdocs-numtide/pull/3

zimbatm:
I'm not sure. @Mic92 do you see any reason not to have this?

SuperSandro2000:
From my personal experience this matches other distros and should be a default in nixos itself actually.

Mic92:
Ok with me

zimbatm:
:+1: 
bors merge

PR #174, opened 2023-06-26
common: don't log refused network connections

phaer:
Hello! 

I think not logging each and every refused connection is a good default, as running a public server quickly leads to unreadable dmesg output, due to scanning et al.


phaer:
CI is supposedly broken because of https://github.com/numtide/mkdocs-numtide/pull/3

zimbatm:
good idea.

bors merge

PR #185, opened 2023-07-08
server: add git to systemPackages

zowoq:
This is useful to have around on hosts.

### DIFF
--- a/nixos/server/default.nix
+++ b/nixos/server/default.nix
@@ -8,7 +8,7 @@
   ];
 
   # List packages installed in system profile.
-  environment.systemPackages = [
+  environment.systemPackages = map lib.lowPrio [
     pkgs.curl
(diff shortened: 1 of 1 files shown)


nixos/server/default.nix
```diff
@@ -11,6 +11,7 @@
   environment.systemPackages = [
     pkgs.curl
     pkgs.dnsutils
+    pkgs.git
```
zimbatm, line 14: how about using gitMinimal so it doesn't increase the closure size too much?

zowoq: Minimal would work but sometimes would want pkgs.git in systemPackages. I've set lowPrio for all of these packages, guess it's the closest to mkDefault?

zimbatm:
bors merge

PR #196, opened 2023-07-16
simplify nixpkgs version checks

Mic92:
Let's make good examples for people to copy. This is considerably shorter and also puts the new thing first


Mic92:
@mergify queue

PR #200, opened 2023-07-17
server: only allow normal users to the nix-daemon

zimbatm:
Reduce the attack surface a little bit.

Just an idea I had while talking to @jfroche

### DIFF
--- a/nixos/server/default.nix
+++ b/nixos/server/default.nix
@@ -37,6 +37,10 @@
   # Delegate the hostname setting to dhcp/cloud-init by default
   networking.hostName = lib.mkDefault "";
 
+  # Only give normal users access to the nix-daemon.
+  nix.settings.allowed-users =
+    map (u: u.name) (lib.filterAttrs (_: u: u.isNormalUser) config.users.users);
+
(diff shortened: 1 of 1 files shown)


zimbatm:
This probably requires a bunch of fixes to the github action runners and queued-build-hook service

SuperSandro2000:
> This probably requires a bunch of fixes to the github action runners and queued-build-hook service

and harmonia

Kranzes:
I implemented something similar in this commit https://github.com/Kranzes/nix-config/commit/81f7269304bd85dd2f473f9fdc0f00d874c1567f. This didn't break `hercules-ci-agent` because it is set in `trusted-users` which seems to bypass `allowed-users`.

Mic92:
@mergify queue

Mic92:
@mergify rebase

zimbatm:
I think this PR is going to break too many things, not sure if it's worth it.

Mic92:
Let's leave this for projects to set.

Kranzes:
> Let's leave this for projects to set.

At least have the function in the diff be correct :^|

PR #208, opened 2023-07-28
common: pin to nix to v2.16 for now

zimbatm:
Nixpkgs distributed nix 2.15 which breaks on nix-copy-closure.


zowoq:
This can be reverted, nix 2.17 is the now default in nixpkgs.

PR #218, opened 2023-08-14
add prometheus module

Mic92:


nixos/roles/prometheus/default.nix
```diff
@@ -0,0 +1,144 @@
...
+            default = { };
+          };
+        };
+      }));
+      example = {
+        prometheusAlerts = {
+          alertRules = {
+            ExampleAlert = {
+              expr = "up == 0";
+              for = "2m";
```
Mic92, line 109:
```suggestion
              for = "4m";
```

nixos/roles/prometheus/default.nix
```diff
@@ -0,0 +1,144 @@
...
+              annotations = {
+                summary = "Instance {{ $labels.instance }} down";
+                description = "{{ $labels.instance }} of job {{ $labels.job }} has been down for more than 2 minutes.";
+              };
+            };
+          };
+          recordingRules = {
+            RecordingExample = {
+              expr = "up";
+              for = "2m";
```
Mic92, line 122:
```suggestion
```

nixos/roles/prometheus/default.nix
```diff
@@ -0,0 +1,144 @@
...
+              labels = {
+                severity = "critical";
+              };
```
Mic92, line 125:
```suggestion
```

nixos/roles/prometheus/default.nix
```diff
@@ -0,0 +1,139 @@
...
+        };
+      };
+    };
+  };
+  config = {
+    services.prometheus = {
+      enable = lib.mkDefault true;
+      # checks fail because of missing secrets in the sandbox
+      checkConfig = false;
+      ruleFiles = [ (pkgs.writers.writeYAML "rules.yaml" rules) ];
```
soupglasses, line 136: Following the documentation, you may set it to `"syntax-only"` to still have some checking.

> If you use credentials stored in external files (password_file, bearer_token_file, etc), they will not be visible to promtool and it will report errors, despite a correct configuration. To resolve this, you may set this option to "syntax-only" in order to only syntax check the Prometheus configuration.
> https://github.com/NixOS/nixpkgs/blob/475d5ae2c4cb87b904545bdb547af05681198fcc/nixos/modules/services/monitoring/prometheus/default.nix#L1738-L1743

nixos/roles/prometheus/default-alerts.nix
```diff
@@ -0,0 +1,229 @@
...
+        annotations.description = "{{$labels.host}}: {{$labels.name}} was not run in the last 24h";
+      };
+
+      TaskFailed = {
+        expr = ''task_last_run{state="fail"}'';
+        annotations.description = "{{$labels.host}}: {{$labels.name}} failed to run";
+      };
+
+      NixpkgsOutOfDate = {
+        expr = ''(time() - flake_input_last_modified{input="nixpkgs",host!="matchbox"}) / (60*60*24) > 7'';
```
phaer, line 70: I think the exclusion of a host named "matchbox" is probably copied from nix-community infra, but not needed for srvos?

nixos/roles/prometheus/default-alerts.nix
```diff
@@ -0,0 +1,229 @@
...
+        annotations.description = "{{$labels.instance}} reports: zpool {{$labels.name}} has {{$value}} errors";
+      };
+
+      MdRaidDegradedDisks = {
+        expr = "mdstat_degraded_disks > 0";
+        annotations.description = "{{$labels.instance}}: raid {{$labels.dev}} has failed disks";
+      };
+
+      # ignore devices that disabled S.M.A.R.T (example if attached via USB)
+      # Also ignore nix-community server ci server until nvme actually fails
```
phaer, line 187: This comment does not seem to be accurate anymore

nixos/roles/prometheus/default-alerts.nix
```diff
@@ -0,0 +1,229 @@
...
+      };
+
+      RamUsing95Percent = {
+        expr = "mem_buffered + mem_free + mem_cached < mem_total * 0.05";
+        for = "1h";
+        annotations.description = "{{$labels.host}} is using at least 95% of its RAM for at least 1 hour";
+      };
+
+      Load15 = {
+        expr = ''system_load15 / system_n_cpus{org!="nix-community"} >= 2.0'';
```
phaer, line 99: nix-community filter included

nixos/roles/prometheus/default.nix
Mic92: applied.

PR #241, opened 2023-09-07
nix-remote-builder: add a proper group and make it a system user

Mic92:
Disclaimer: this is untested, I just saw the these things and fixed them.

- normal users show up in login manager, we don't want this for this user
- nogroup shouldn't be used for anything otherwise it becomes somegroup (i.e. multiple users could share the same group)

### DIFF
--- a/nixos/roles/nix-remote-builder.nix
+++ b/nixos/roles/nix-remote-builder.nix
@@ -26,14 +26,18 @@ in
       { domain = "*"; item = "nofile"; type = "-"; value = "20480"; }
     ];
 
-    # Give restricted SSH access to the build scheduler
-    users.users.nix-remote-builder.openssh.authorizedKeys.keys = map
-      (key:
-        ''command="nix-daemon --stdio",no-agent-forwarding,no-port-forwarding,no-pty,no-user-rc,no-X11-forwarding ${key}''
-      )
-      cfg.schedulerPublicKeys;
-    users.users.nix-remote-builder.isNormalUser = true;
-    users.users.nix-remote-builder.group = "nogroup";
+    users.users.nix-remote-builder = {
+      isSystemUser = true;
+      group = "nix-remote-builder";
+      # Give restricted SSH access to the build scheduler
+      openssh.authorizedKeys.keys = map
+        (key:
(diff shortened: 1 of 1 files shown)


Mic92:
I wonder why the GC is so aggressive on this module. I wouldn't not expect such a module to set any gc at all, so it can be deployed on ci servers.

nixos/roles/nix-remote-builder.nix
```diff
@@ -26,14 +26,18 @@ in
       { domain = "*"; item = "nofile"; type = "-"; value = "20480"; }
```
Mic92, line 26: Does this even have any effect for the nix-daemon service that hosts the builds?

nixos/roles/nix-remote-builder.nix
```diff
@@ -26,14 +26,18 @@ in
...
-    # Give restricted SSH access to the build scheduler
-    users.users.nix-remote-builder.openssh.authorizedKeys.keys = map
-      (key:
-        ''command="nix-daemon --stdio",no-agent-forwarding,no-port-forwarding,no-pty,no-user-rc,no-X11-forwarding ${key}''
-      )
-      cfg.schedulerPublicKeys;
-    users.users.nix-remote-builder.isNormalUser = true;
-    users.users.nix-remote-builder.group = "nogroup";
+    users.users.nix-remote-builder = {
+      isSystemUser = true;
```
zowoq, line 30: `isSystemUser` sets the shell to `nologin`. Might be simpler to just keep isNormalUser instead?

Mic92: Normal Users show up in login manager. I have to check what heuristic this is based on. But would be nice if service accounts don't show up in there. Also having a group not equal to users is desirable since this may have other side effects.

Mic92: It's a common pattern to use the normal user flag as a filter for other things in the module system.

zowoq: > It's a common pattern to use the normal user flag as a filter for other things in the module system.

This is sort of what I meant, it depends on the filter, shell and ssh isn't something system users would normally have.

PR #250, opened 2023-09-19
hetzner-online: use predictable interface names

phaer:
...and match probable names of the first interface according to
https://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/

matchConfig.Name is a whitespace-separated list of shell-style globs according to man 5 systemd.network.

old-style eth0 is kept to ease switching between interface naming schemes and because it seems harmless to do so.

### DIFF
--- a/nixos/hardware/hetzner-online/default.nix
+++ b/nixos/hardware/hetzner-online/default.nix
@@ -32,7 +32,7 @@
     networking.useDHCP = false;
 
     systemd.network.networks."10-uplink" = {
-      matchConfig.Name = "eno1 ens1 enp1s0 eth0";
+      matchConfig.Name = lib.mkDefault "eno1 ens1 enp1s0 eth0";
       networkConfig.DHCP = "ipv4";
       # hetzner requires static ipv6 addresses
       networkConfig.Gateway = "fe80::1";
--- a/nixos/hardware/hetzner-online/default.nix
+++ b/nixos/hardware/hetzner-online/default.nix
@@ -32,7 +32,7 @@
     networking.useDHCP = false;
 
     systemd.network.networks."10-uplink" = {
-      matchConfig.Name = lib.mkDefault "eno1 ens1 enp1s0 eth0";
+      matchConfig.Name = lib.mkDefault "en* eth0";
       networkConfig.DHCP = "ipv4";
       # hetzner requires static ipv6 addresses
       networkConfig.Gateway = "fe80::1";
--- a/nixos/hardware/hetzner-online/default.nix
+++ b/nixos/hardware/hetzner-online/default.nix
@@ -30,11 +30,9 @@
 
     networking.useNetworkd = true;
     networking.useDHCP = false;
-    # Hetzner servers commonly only have one interface, so its either to just match by that.
-    networking.usePredictableInterfaceNames = false;
 
     systemd.network.networks."10-uplink" = {
-      matchConfig.Name = "eth0";
+      matchConfig.Name = "eno1 ens1 enp1s0 eth0";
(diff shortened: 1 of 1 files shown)


nixos/hardware/hetzner-online/default.nix
```diff
@@ -30,11 +30,9 @@
 
     networking.useNetworkd = true;
     networking.useDHCP = false;
-    # Hetzner servers commonly only have one interface, so its either to just match by that.
-    networking.usePredictableInterfaceNames = false;
 
     systemd.network.networks."10-uplink" = {
-      matchConfig.Name = "eth0";
+      matchConfig.Name = "eno1 ens1 enp1s0 eth0";
```
Mic92, line 35: Only question is what we should do if we have multiple interfaces. It's a bit of a special case and maybe we can get away with just documenting it. But if you order a second interface than you end up with two default ipv6 gateways.

I also would than just match for `matchConfig.Type = "ethernet";`.

phaer: > But if you order a second interface than you end up with two default ipv6 gateways.

I think that's only true if e.g. your first interface is onboard and you second isn't so that you end up with e.g. eno1 & enp1s0. Otherwise you would either end up with eno1 & eno2 or e.g. enp2s0 & enp2s0, were only the former would match.

> I also would than just match for matchConfig.Type = "ethernet";.

But that would match *all* ethernet devices, I deliberately only do a (dirty) match for the first one :)

Mic92: I see.

Mic92: Fine with me than.

Mic92 (approved):
Looks good to me merge yourself or someone else who feels comfortable/has tested it.

nixos/hardware/hetzner-online/default.nix
Mic92, line 35: We can make this easier to override than:

```suggestion
      matchConfig.Name = lib.mkDefault "eno1 ens1 enp1s0 eth0";
```

nixos/hardware/hetzner-online/default.nix
```diff
@@ -30,11 +30,9 @@
 
     networking.useNetworkd = true;
     networking.useDHCP = false;
-    # Hetzner servers commonly only have one interface, so its either to just match by that.
-    networking.usePredictableInterfaceNames = false;
 
     systemd.network.networks."10-uplink" = {
-      matchConfig.Name = "eth0";
+      matchConfig.Name = lib.mkDefault "eno1 ens1 enp1s0 eth0";
```
phaer, line 35: > enp1s0

Upon revisiting this, i think this part might be to naive. Not sure if enp0s0 would be better, need to check docs again.

Mic92: https://www.freedesktop.org/software/systemd/man/systemd.net-naming-scheme.html

Mic92:
Have you had a look at this again?

nixos/hardware/hetzner-online/default.nix
phaer: My question above would only have been relevant if we needed a consistent definition of the "first" interface. We don't need that here, and so we can just match on all ethernet devices.

phaer:
> Have you had a look at this again?

Just did, but couldn't self-merge even if i wanted to due to lack of repo permissions.

phaer:
Ah, I did gain the right to self-merge since my last post :D 

Doing so this time as we've discussed this in the past, I am running it on multiple hetzner machines and we could revert it if it turns out to break anyones setup.

PR #257, opened 2023-09-26
telegraf: add internal

zowoq:
Useful to see drops/errors/timeouts from the agent.


nixos/mixins/telegraf.nix
```diff
@@ -129,6 +129,7 @@ in
           device = [ "rpc_pipefs" "lxcfs" "nsfs" "borgfs" ];
         };
         diskio = { };
+        internal = { };
```
Mic92, line 132: Do you got some interesting alert rules that I could add here as well?

zowoq: I was thinking that these could be displayed on grafana but we could write rules as well.

PR #264, opened 2023-10-06
nix-daemon: enable more experimental features

Mic92:
Impure-derivations and auto-allocate-uids seem to be pretty stable and are not used if not requested by a build. repl-flake changes the behaviour of nix repl, but this change will be considered stable in the next release anyway.

### DIFF
--- a/nixos/common/nix.nix
+++ b/nixos/common/nix.nix
@@ -7,6 +7,9 @@
   nix.settings.experimental-features = [
     "nix-command"
     "flakes"
+    "repl-flake"
+    "impure-derivations"
+    "auto-allocate-uids"
   ];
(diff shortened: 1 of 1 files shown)


phaer (approved):

nixos/common/nix.nix
```diff
@@ -7,6 +7,9 @@
   nix.settings.experimental-features = [
     "nix-command"
     "flakes"
+    "repl-flake"
+    "impure-derivations"
+    "auto-allocate-uids"
```
Kranzes, line 12: for `auto-allocate-uids` to work you need to set `settings.auto-allocate-uids = true;`

PR #269, opened 2023-10-13
nix-daemon: no longer preallocate users and enable uid-range feature …

Mic92:
…for container builds

### DIFF
--- a/nixos/common/nix.nix
+++ b/nixos/common/nix.nix
@@ -8,8 +8,6 @@
     "nix-command"
     "flakes"
     "repl-flake"
-    "impure-derivations"
-    "auto-allocate-uids"
   ];
 
   # The default at 10 is rarely enough.
--- a/docs/nixos/mixins.md
+++ b/docs/nixos/mixins.md
@@ -14,9 +14,14 @@ Configure systemd-boot as bootloader.
 
 Enables a generic telegraf configuration. See [Mic's dotfiles](https://github.com/Mic92/dotfiles/blob/master/nixos/eva/modules/prometheus/alert-rules.nix) for monitoring rules targeting this telegraf configuration.
 
-### `nixosModules.nginx`
+### `nixosModules.mixins-nginx`
 
 Configure Nginx with recommended settings. Is quite useful when using nginx as a reverse-proxy on the machine to other services.
+
+### `nixosModules.mixins-nix-experimental`
+
+Enables all experimental features in nix, that are known safe to use (i.e. are only used when explicitly requested in a build).
+This for example unlocks use of containers in the nix sandbox.
 
 ### `nixosModules.mixins-trusted-nix-caches`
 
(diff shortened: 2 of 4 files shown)


zimbatm:
how reliable is this?

zowoq:
These should be opt in, not enabled by default.

nixos/common/nix.nix
```diff
@@ -9,9 +9,17 @@
     "flakes"
     "repl-flake"
     "impure-derivations"
+    # for container in builds support
     "auto-allocate-uids"
+    "cgroup"
   ];
 
+  # no longer need to pre-allocate build users for everything
+  nix.settings.auto-allocate-uids = true;
```
SuperSandro2000, line 18: Can I ask a stupid question: Would I need to set use-cgroups, too? When running nix show-config it showed as being disabled but auto-allocate-uids mentions that it requires this feature and it worked without it for me.

Mic92:
> These should be opt in, not enabled by default.

Please explain. I missing the context here.

Mic92:
> how reliable is this? It has the potential to break a lot of things.

You have to explicitly request this in your derivation: https://github.com/NixOS/nix/blob/955bbe53c5e7e742f7f584c54f54b09fd5ce4009/tests/nixos/containers/systemd-nspawn.nix#L47
Otherwise it's a no-op.

zowoq:
> Regarding `nix.settings.auto-allocate-uids`: This should have been tested in production by the determinate systems installer for a while.

It was disabled last month.

phaer:
@zowoq If we are talking about https://github.com/DeterminateSystems/nix-installer/pull/638 et al: This seems to affect non-nixos linux distros only, while srvos only runs on nixos?

zowoq:
I wasn't talking about why, I was just pointing out that it isn't currently enabled.

phaer:
Okay, so the answer to the question why those settings should not be enabled by default is left as an exercise to the reader? 😄

zowoq:
> Okay, so the answer to the question why those settings should not be enabled by default is left as an exercise to the reader? 😄

I don't understand? I was responding to this:

> This should have been tested in production by the determinate systems installer for a while.

It was is only "tested in production" for a few months.

phaer:
Maybe I misread, sorry for the noise if so. In my reading the conversation had started with 

> these should be opt in, not enabled by default

Which sounded like a response implying specific knowledge about the question in the post before it. Anyway, it's merged so we'll gonna find out 😄

Mic92:
> Maybe I misread, sorry for the noise if so. In my reading the conversation had started with
> 
> > these should be opt in, not enabled by default
> 
> Which sounded like a response implying specific knowledge about the question in the post before it. Anyway, it's merged so we'll gonna find out 😄

It's not enabled by default in srvos. It's a different module that can be enabled optionally.

PR #285, opened 2023-11-04
Disable sudo lecture

pingiun:
Whether the sudo lecture is shown is stateful. On nixos systems that don't store state you will see the sudo lecture a lot, so I find it useful to disable it

### DIFF
--- a/nixos/common/default.nix
+++ b/nixos/common/default.nix
@@ -38,6 +38,9 @@
 
   # Allow sudo from the @wheel group
   security.sudo.enable = true;
+  security.sudo.extraConfig = ''
+    Defaults lecture = never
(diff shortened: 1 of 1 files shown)


zimbatm:
Sounds good 👍

Mic92:
Sounds good to me.

SuperSandro2000:
This is just a problem for people that don't store state which is the wast minority.

pingiun:
> This is just a problem for people that don't store state which is the wast minority.

On servers I think the amount would be higher, but in any case I don't think it's a bad default. The sudo lecture doesn't really add anything imo

phaer (approved):

PR #289, opened 2023-11-05
sudo: only allow exec by wheel

zimbatm:
Was found in https://xeiaso.net/blog/paranoid-nixos-2021-07-18/

I'm also moving things out of the default.nix, so they can more easily be cherry-picked.

### DIFF
--- a/nixos/common/default.nix
+++ b/nixos/common/default.nix
@@ -10,6 +10,7 @@
     ./nix.nix
     ./openssh.nix
     ./serial.nix
+    ./sudo.nix
     ./upgrade-diff.nix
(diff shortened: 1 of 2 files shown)


phaer:
sudo.nix seems to be missing from this PR?

zimbatm:
good catch :)

phaer (approved):
Works as designed for me and I see no drawbacks to it.

CI checks seem to indefinitely(?) wait for a runner atm, but I think @zimbatm is still working on that?

SuperSandro2000 (approved):
I have turned this option on in my private config since December 2021 and so far didn't encounter any issues.

PR #290, opened 2023-11-05
openssh: disable SFTP by default

zimbatm:
Also inspired by https://xeiaso.net/blog/paranoid-nixos-2021-07-18/

### DIFF
--- a/nixos/common/openssh.nix
+++ b/nixos/common/openssh.nix
@@ -2,10 +2,12 @@
 { config, lib, ... }:
 {
   services.openssh = {
-    settings.X11Forwarding = false;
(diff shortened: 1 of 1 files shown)


Mic92:
I feel like I would use this feature too often to have it disabled and an attacker that has ssh access also can extract files, so I don't see a significant security win.

zimbatm:
Ok

zimbatm:
Just to be sure; this is SFTP, not scp

phaer:
`scp`, the tool, uses SFTP under the hood since [OpenSSH 9.0](https://www.openssh.com/txt/release-9.0). Disabling SFTP would require use of the legacy scp protocol

PR #291, opened 2023-11-05
ci: restore GitHub Actions

zimbatm:
This is a subset that should work (it's going to be slow though)

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -9,12 +9,8 @@ on:
   workflow_dispatch:
 jobs:
   build:
-    strategy:
-      matrix:
-        system:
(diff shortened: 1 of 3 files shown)


zimbatm:
I see the namespace.so ran out of memory: https://cloud.namespace.so/4l6g3pb71m64u/cluster/rfn8s20s5hh78
image

zimbatm:
the server VM test is super slow

Mic92:
We can just expose our nixos tests in `.#checks` for now and leave the rest for the github CI.
I am thinking about how I can make buildbot reading a configuration file to make build outputs and subdirs configurable.

zimbatm:
good idea

PR #295, opened 2023-11-06
github: fix settings

zimbatm:


.github/settings.yml
```diff
@@ -123,15 +123,15 @@ branches:
       # Required. Require status checks to pass before merging. Set to null to disable
       required_status_checks:
         # Required. Require branches to be up to date before merging.
-        strict: true
+        strict: false
         # Required. The list of status checks to require in order to merge into this branch
-        contexts: [ "bors" ]
```
zimbatm, line 128: I'm not sure what to put here

zowoq: Could probably leave this as null for now and revisit it once CI has been stable for a bit.

.github/settings.yml
```diff
@@ -123,15 +123,15 @@ branches:
...
+        contexts: []
       # Required. Enforce all configured restrictions for administrators. Set to true to enforce required status checks for repository administrators. Set to null to disable.
       enforce_admins: false
       # Disabled for bors to work
       required_linear_history: false
       # Required. Restrict who can push to this branch. Team and user restrictions are only available for organization-owned repositories. Set to null to disable.
       restrictions:
-        apps: [ "bors" ]
+        apps: []
```
zowoq, line 135: I think mergify needs permissions like bors.

```suggestion
        apps: [ "mergify" ]
```

zimbatm: This means that only mergify can push to main. Is that what we want?

.github/settings.yml
```diff
@@ -123,15 +123,15 @@ branches:
...
         users: []
         teams: []
```
zowoq, line 137:
```suggestion
        teams: [ "srvos" ]
```

PR #297, opened 2023-11-07
prometheus: various fixes

zowoq:


nixos/roles/prometheus/default-alerts.nix
```diff
@@ -115,18 +115,18 @@
       };
 
       TelegrafDown = {
-        expr = ''min(up{job=~"telegraf",type!='mobile'}) by (source, job, instance, org) == 0'';
```
Mic92: Isn't this mobile filter a quite generic thing? I am using this for laptop monitoring or in general devices that are allowed to go offline.

nixos/roles/prometheus/default-alerts.nix
```diff
@@ -75,7 +75,7 @@
       };
 
       SwapUsing30Percent = {
-        expr = ''mem_swap_total{host!="eva"} - (mem_swap_cached + mem_swap_free) > mem_swap_total * 0.3'';
```
Mic92, line 78: Fine with me. This host in question doesn't even swap anymore.

Mic92 (approved):

nixos/roles/prometheus/default-alerts.nix
```diff
@@ -115,18 +115,18 @@
       };
 
       TelegrafDown = {
-        expr = "min(up{job=~\"telegraf\",type!='mobile'}) by (source, job, instance, org) == 0";
```
Mic92, line 118: Isn't this mobile filter a quite generic thing? I am using this for laptop monitoring or in general devices that are allowed to go offline.

zowoq: I've reverted this but I don't think they belong in this repo.

PR #304, opened 2023-11-12
sudo: drop explicit sudo enable

Mic92:
This makes it harder to replace sudo with sudo-rs


zimbatm:
In that case sudo won't be enabled isn't it?

Mic92:
Yes. It's [security.sudo-rs.enable](https://search.nixos.org/options?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=sudo-rs). Sudo will be still enabled by default, so nothing srvos needs to change - if I user does disable sudo they have a reason to.

PR #312, opened 2023-11-19
boot.initrd.systemd: also enable if initrd networkd is used

Mic92:
fixes https://github.com/nix-community/srvos/pull/312


Mic92:
@mergify queue

PR #330, opened 2023-12-09
hardware-hetzner-cloud-arm: enable early console

elohmeier:
This fixes the console being unavailable during early boot on Hetzner Cloud ARM servers.
It is required to be able to enter a LUKS password during boot.


Mic92:
@mergify queue

PR #333, opened 2023-12-11
openssh: don't force authorizedKeysFiles if forgejo is enabled

priegger:
Forgejo is a fork of gitea and needs the same exception.


Mic92:
@mergify queue

Mic92:
@mergify refresh

Mic92:
@mergify queue

PR #350, opened 2024-01-01
dev flake

zowoq:
This allows us to have private dev inputs that we can use in the main flake and move the github actions checks to buildbot.

We may be able to upstream the vendored files to the flake-parts repo but I don't want bother with that for now.

~~Depends on https://github.com/nix-community/srvos/pull/355.~~


zimbatm (changes requested):
What problem does it solve? Please add more details to your commit messages so it's possible to follow the reasoning.

zowoq:
It allows us to have private dev inputs that we can use in the main flake, can then move the github actions checks to buildbot. We discussed this briefly on matrix a few weeks ago.

zowoq:
I plan to merge this before the next flake update.

.github/workflows/update-flake-lock.yml
```diff
@@ -12,9 +12,25 @@ jobs:
...
+          pushd dev/private
+          nix flake update --commit-lock-file
+          popd
+          nix run .#update-dev-private-narHash
+          git commit --amend --no-edit dev/private.narHash
+      - uses: peter-evans/create-pull-request@v5
+        # disable for now
+        #with:
+        #  labels: |
+        #    merge-queue
```
zowoq, line 36: I'll merge these manually for now, will re-enable this in a week or two.

PR #351, opened 2024-01-02
only enable configurable-impure-env in nix 2.19 or newer

Mic92:


Mic92:
@mergify queue

Mic92:
@mergify queue

PR #354, opened 2024-01-04
add hardware-hetzner-online-arm

zowoq:


nixos/default.nix
```diff
@@ -44,6 +44,7 @@ exposeModules ./. [
   ./hardware/hetzner-online/amd.nix
   ./hardware/hetzner-online/intel.nix
   ./hardware/hetzner-online/ex101.nix
+  ./hardware/hetzner-online/rx170.nix
```
Mic92, line 47: I would generalize this, since we got two models that are basically the same SOC with different hardware builtin:

```suggestion
  ./hardware/hetzner-online/arm.nix
```

PR #383, opened 2024-02-13
no stub: <insert TLC song>

zimbatm:
This feature was introduced in
https://github.com/NixOS/nixpkgs/pull/269551 and is mostly useful for
servers.

Saves around 80MiB of evaluation on x86_64 due to the extra i686 nixpkgs
instance.

### DIFF
--- a/nixos/server/default.nix
+++ b/nixos/server/default.nix
@@ -1,89 +1,95 @@
 # A default configuration that applies to all servers.
 # Common configuration accross *all* the machines
-{ pkgs, lib, ... }:
+{ options, pkgs, lib, ... }:
 {
 
   imports = [
     ../common
   ];
 
-  # List packages installed in system profile.
-  environment.systemPackages = map lib.lowPrio [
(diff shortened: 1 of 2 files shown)


nixos/server/default.nix
```diff
@@ -1,89 +1,95 @@
...
-      # it will forcefully reboot the system.
-      runtimeTime = "20s";
-      # Forcefully reboot if the final stage of the reboot
-      # hangs without progress for more than 30s.
-      # For more info, see:
-      #   https://utcc.utoronto.ca/~cks/space/blog/linux/SystemdShutdownWatchdog
-      rebootTime = "30s";
+  config = {
+    # List packages installed in system profile.
+    environment.systemPackages = map lib.lowPrio [
```
zowoq, line 12: Since this change these packages aren't installed?

zimbatm: I just checked and it looks like it's the case, but I'm unsure why.

PR #406, opened 2024-04-02
hetzner-cloud/arm: remove boot.loader.timeout

elohmeier:
This timeout seems to be quite arbitrary, it is not specifically required for Hetzner ARM VMs.


Mic92:
@mergify queue

Mic92:
@mergify rebase

Mic92:
@mergify queue

PR #412, opened 2024-04-10
add digitalocean

Mic92:
untested but maybe @bbigras has already some thoughts on it.

### DIFF
--- a/docs/nixos/hardware.md
+++ b/docs/nixos/hardware.md
@@ -10,6 +10,12 @@ Hardware configuration for <https://aws.amazon.com/ec2> instances.
 
 The main difference here is that the default userdata service is replaced by cloud-init.
 
+### `nixosModules.hardware-digitalocean`
+
+Hardware configuration for <https://www.digitalocean.com/> instances.
+
+Enables cloud-init but turns of non-working dhcp.
+
 ### `nixosModules.hardware-hetzner-cloud`
 
 Hardware configuration for <https://www.hetzner.com/cloud> instances.
--- a/docs/nixos/hardware.md
+++ b/docs/nixos/hardware.md
@@ -10,7 +10,7 @@ Hardware configuration for <https://aws.amazon.com/ec2> instances.
 
 The main difference here is that the default userdata service is replaced by cloud-init.
 
-### `nixosModules.hardware-digitalocean`
+### `nixosModules.hardware-digitalocean-droplet`
 
 Hardware configuration for <https://www.digitalocean.com/> instances.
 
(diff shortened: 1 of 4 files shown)


docs/nixos/hardware.md
```diff
@@ -10,6 +10,12 @@ Hardware configuration for <https://aws.amazon.com/ec2> instances.
 
 The main difference here is that the default userdata service is replaced by cloud-init.
 
+### `nixosModules.hardware-digitalocean`
```
bbigras, line 13:
```suggestion
### `nixosModules.hardware-digitalocean-droplet`
```

bbigras:
I was able to deploy a working droplet using this. The normal digitalocean console doesn't work, but the recovery one does and I can ssh in.

If I don't use this module and I only import `(modulesPath + "/virtualisation/digital-ocean-config.nix")` I can't ssh in. The recovery console still works, though.

If someone wants to test, here's a terraform module to create a droplet and install nixos on it with nixos-anywhere:
https://gitlab.com/bbigras/test-srvos-digitalocean

dfrankland:
I'm a simple bystander, but I happened to be trying to deploy from darwin-arm64 to a digitalocean droplet today. [test-srvos-digitalocean](https://gitlab.com/bbigras/test-srvos-digitalocean) was the only example I could find that actually worked and allowed SSH after finishing a deploy with `nixos-anywhere`

bbigras:
@dfrankland I'm glad that it was useful to you :). Note that you can use `ipv6 = true` with `digitalocean_droplet` too.

I think I was using `modules/virtualisation/digital-ocean-image.nix` before and IIRC, digital ocean doesn't support ipv6 with custom images.

bbigras:
Should we set `systemd.enableEmergencyMode = true`?

https://github.com/nix-community/srvos/blob/e4252aa777482dc9d4cacd779ae29115de69b7ba/nixos/server/default.nix#L62C1-L66C33

Mic92:
Does the normal digitalocean console uses some daemon? As of know we don't deploy any ssh keys that digitalocean may try to add for the console to work.

Mic92:
@mergify queue

bbigras:
> Does the normal digitalocean console uses some daemon? As of know we don't deploy any ssh keys that digitalocean may try to add for the console to work.

image

when you connect, you don't see a login prompt:

image

Mic92:
Ok. So we would allow dynamic ssh keys for this one.

bbigras:
I just tested with `virtualisation.digitalOcean.setSshKeys = true;` and it doesn't seem to be enough.

I don't see anything in `journalctl -f -u sshd`.

Mic92:
It's because we don't allow keys outside of `/etc/ssh/authorized_keys.d/`. See this: https://github.com/nix-community/srvos/blob/937ddb11d81d9706b26dc583cf41e65de771c346/nixos/common/openssh.nix#L32

PR #415, opened 2024-04-12
enable multicast dns

Mic92:
It's quite cool if you can just ping machines by hostname.

### DIFF
--- a/nixos/default.nix
+++ b/nixos/default.nix
@@ -56,6 +56,7 @@ exposeModules ./. [
   ./mixins/tracing.nix
   ./mixins/trusted-nix-caches.nix
   ./mixins/nix-experimental.nix
+  ./mixins/mdns.nix
   ./roles/github-actions-runner.nix
   ./roles/nix-remote-builder.nix
   ./roles/prometheus
--- a/nixos/desktop/default.nix
+++ b/nixos/desktop/default.nix
@@ -1,6 +1,7 @@
 { lib, ... }: {
   imports = [
     ../common
+    ../mixins/mdns.nix
     ./pipewire.nix
   ];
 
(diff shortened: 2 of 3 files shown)


nixos/common/networking.nix
```diff
@@ -24,4 +24,14 @@
   systemd.services.systemd-networkd.stopIfChanged = false;
   # Services that are only restarted might be not able to resolve when resolved is stopped before
   systemd.services.systemd-resolved.stopIfChanged = false;
+
+  # Allows to find machines on the local network by name
+  systemd.network.networks."99-ethernet-default-dhcp".networkConfig = {
+    MulticastDNS = lib.mkDefault "yes";
+    LLMNR = lib.mkDefault "yes";
```
SuperSandro2000, line 31: LLMNR is being deprecated and even Microsoft recommends disabling it.

Mic92: I didn't know that Windows does now mdns as well.

zimbatm:
I'm unsure about the implications of this change since it opens the machine to new attack vectors. Maybe move it to mixins?

nixos/common/networking.nix
```diff
@@ -24,4 +24,9 @@
   systemd.services.systemd-networkd.stopIfChanged = false;
   # Services that are only restarted might be not able to resolve when resolved is stopped before
   systemd.services.systemd-resolved.stopIfChanged = false;
+
+  # Allows to find machines on the local network by name
+  systemd.network.networks."99-ethernet-default-dhcp".networkConfig.MulticastDNS = lib.mkDefault "yes";
+  systemd.network.networks."99-wireless-client-dhcp".networkConfig.MulticastDNS = lib.mkDefault "yes";
+  networking.firewall.allowedUDPPorts = [ 5353 ]; # Multicast DNS
```
Mic92, line 31: I also learned today that without this rule, you cannot resolve other mdns names as well. I.e. many printers these days only use mdns for discovery, which works on Ubuntu but breaks on NixOS.

Mic92:
I moved it now to the desktop profile.

nixos/common/networking.nix
SuperSandro2000: Hence avahi opens it by default because otherwise it is quite literal useless but pipe wire only works with avahi, so I kinda need to stick to it.

Mic92: Would avahi create a conflict with these rules? We can also check for avahi here.

PR #431, opened 2024-05-28
common: use new rust-based switch script

Mic92:
This implementation is meant to replace switch-to-configuration.pl with a rust version.
I tested it on my infra for now.

### DIFF
--- a/nixos/common/default.nix
+++ b/nixos/common/default.nix
@@ -1,6 +1,6 @@
 # A default configuration that applies to all servers.
 # Common configuration across *all* the machines
-{ config, lib, ... }:
+{ options, config, lib, ... }:
 {
 
   imports = [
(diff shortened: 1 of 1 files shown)


zimbatm:
Looks like it's not available on aarch64 yet?

Mic92:
Fixed, but we need to disable it on stable.

nixos/common/default.nix
```diff
@@ -16,6 +16,13 @@
     ./zfs.nix
   ];
 
+  system.switch = {
+    enable = lib.mkDefault false;
```
zimbatm, line 20: @Mic92 what happens when this is set to false and enableNg is unset?

SuperSandro2000: Then you can't switch. That's what the perlless tests are using.

Mic92: Ah hang on. This is actually pretty bad if you are on an old version.

Mic92: Fixed in https://github.com/nix-community/srvos/pull/433/files

PR #440, opened 2024-06-13
sync boot.initrd.systemd.network with config.systemd.network

Mic92:
cc @SuperSandro2000

### DIFF
--- a/nixos/common/networking.nix
+++ b/nixos/common/networking.nix
@@ -1,4 +1,4 @@
-{ lib, ... }:
+{ config, lib, ... }:
 {
   # Allow PMTU / DHCP
   networking.firewall.allowPing = true;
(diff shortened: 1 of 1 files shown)


nixos/common/networking.nix
```diff
@@ -24,4 +24,9 @@
   systemd.services.systemd-networkd.stopIfChanged = false;
   # Services that are only restarted might be not able to resolve when resolved is stopped before
   systemd.services.systemd-resolved.stopIfChanged = false;
+
+  # If networkd in initrd is enabled, just copy the configuration of the normal networkd.
+  # Potentially can miss some drivers, but hopefully it would gracefully degrade in those cases
+  # See https://github.com/NixOS/nixpkgs/issues/157034#issuecomment-2163512812
+  boot.initrd.systemd.network = lib.mkDefault (builtins.removeAttrs config.systemd.network ["enable"]);
```
Mic92, line 31: Also I am unsure if this is a good idea. If for example someone sets up a wireguard tunnel, they will likely miss the keys in initrd. There is probably more examples like that.

SuperSandro2000: or when doing bridges/bonding you need extra kernel modules https://gitea.c3d2.de/c3d2/nix-config/src/branch/master/modules/baremetal.nix#L18

Mic92:
I put this on the "too much magic" pile.

PR #441, opened 2024-06-13
Darwin support

Mic92:
Now that this seems to be now also part of my own infastructure,
I kind of miss having srvos to share configuration.
Also there is some opportunity to share code with NixOS,
I decided to not do so as it would make contributing to this project harder
as not everyone has access to both Linux and MacOS at the same time.

I think it's acceptable to have a slight code drift, but allow to properly test
changes on both platforms in their own time.

I still need to update documentation here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,8 @@ example to deploy a GitHub Action runner on Hetzner:
 
 The [Documentation](https://nix-community.github.io/srvos/) website shows more general usage, how to install SrvOS, etc...
 
-To improve the documentation, take a look at the `./docs` folder. You can also run `nix run .#docs.serve` to start a preview server on <http://localhost:3000>.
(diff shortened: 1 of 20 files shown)


zowoq:
> Also there is some opportunity to share code with NixOS, I decided to not do so

> I think it's acceptable to have a slight code drift, but allow to properly test changes on both platforms in their own time.

I wanted to share code and avoid drift so this is the opposite of what I wanted to do here.

If this is what you want to do, okay, I'm not interested in debating it but I'm not going to use it for the community infra.

zimbatm:
Sounds good.

To have shared code, we need CI to exercise the Darwin code first. The CI is pretty fast so I think it would be reasonable after that.

Mic92:
@mergify queue

PR #449, opened 2024-07-01
common/openssh: apply workaround for CVE-2024-6387

Mic92:


Mic92:
Works on my machines (tm)

emilazy:
The fix is now in all the channels, so I’d recommend removing this mitigation to avoid the DoS vulnerability it opens up.

zowoq:
Reverted in https://github.com/nix-community/srvos/pull/451.

PR #455, opened 2024-07-08
hetzner-cloud: enable qemu-agent for password resets

zimbatm:
This is useful to log in using the VNC console.

Tested, works as expected


nixos/hardware/hetzner-cloud/default.nix
```diff
@@ -13,5 +13,10 @@
 
     networking.useNetworkd = true;
     networking.useDHCP = false;
+
+    # Needed by the Hetzner Cloud password reset feature.
+    services.qemuGuest.enable = lib.mkDefault true;
```
Mic92, line 18: I think this is worth mentioning in documentation. Otherwise looks good to me.

PR #463, opened 2024-07-15
feat(server): restrict boot entries to 5

zimbatm:
Avoid the full /boot partition issue by limiting the number of entries to add in said partition.

Since a boot entry (bzImage + initrd) is around 32MiB, this require the /boot partition to be at least 160MiB in size, 260MiB+ recommended (for the block sizes).

Ref: <https://www.ctrl.blog/entry/esp-size-guide.html>

### DIFF
--- a/nixos/mixins/systemd-boot.nix
+++ b/nixos/mixins/systemd-boot.nix
@@ -1,11 +1,11 @@
 # Add this mixin to machines that boot with EFI
+{ lib, ... }:
 {
-
   # Only enable during install
   #boot.loader.efi.canTouchEfiVariables = true;
 
   # Use systemd-boot to boot EFI machines
-  boot.loader.systemd-boot.configurationLimit = 10;
+  boot.loader.systemd-boot.configurationLimit = lib.mkOverride 1337 10;
   boot.loader.systemd-boot.enable = true;
   boot.loader.timeout = 3;
 }
--- a/nixos/server/default.nix
+++ b/nixos/server/default.nix
@@ -27,6 +27,12 @@
     stub-ld.enable = lib.mkDefault false;
   };
 
+  # Restrict the number of boot entries to prevent full /boot partition.
+  #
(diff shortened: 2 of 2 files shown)


nixos/server/default.nix
```diff
@@ -24,6 +24,12 @@
     stub-ld.enable = lib.mkDefault false;
   };
 
+  # Restrict the number of boot entries to prevent full /boot partition.
+  #
+  # Servers don't need too many generations.
+  boot.loader.grub.configurationLimit = lib.mkDefault 5;
+  boot.loader.systemd-boot.configurationLimit = lib.mkDefault 5;
```
Mic92, line 31: Does this also contain the currently booted kernel?

SuperSandro2000: I don't think so

Mic92: I got told the booted kernel will be still protected by /run/booted-system

zimbatm: Ok then good to go?

phaer (commented):
I think lowering the limit for servers is a good idea. But this seems to remove the current limit of 10 generations from non-servers as well?

Mic92:
Sound like a good idea.

nixos/mixins/systemd-boot.nix
```diff
@@ -5,7 +5,6 @@
   #boot.loader.efi.canTouchEfiVariables = true;
 
   # Use systemd-boot to boot EFI machines
-  boot.loader.systemd-boot.configurationLimit = 10;
```
phaer, line 8: I believe we need to either restore this line as a generic default for non-servers or explicitly set it in the desktop profile. Otherwise all non-servers fall back to nixpkgs default of `null` instead of the intended `10` generations.

zimbatm: Yup. Restored to 10 with a lower priority.

PR #464, opened 2024-07-15
nix: enable auto-optimize-store

Mic92:


nixos/common/nix.nix
```diff
@@ -13,6 +13,9 @@
   # The default at 10 is rarely enough.
   nix.settings.log-lines = lib.mkDefault 25;
 
+  # De-duplicate store paths using hardlinks
+  nix.settings.auto-optimise-store = lib.mkDefault true;
```
SuperSandro2000, line 17: In my experience this still slows down copies of store paths significantly. IMO the systemd timer is a better alternative which doesn't slow down every store operation.

Mic92: Makes sense. We can have the timer also on both platforms.

Mic92: Is there an option for this?

SuperSandro2000: yep https://search.nixos.org/options?channel=unstable&show=nix.optimise.automatic&from=0&size=50&sort=relevance&type=packages&query=nix.optimise

Mic92: Applied.

nixos/common/nix.nix
```diff
@@ -13,6 +13,9 @@
   # The default at 10 is rarely enough.
   nix.settings.log-lines = lib.mkDefault 25;
 
+  # De-duplicate store paths using hardlinks
+  nix.optimise.automatic = true;
```
zowoq, line 17:
```suggestion
  nix.optimise.automatic = lib.mkDefault true;
```

Mic92:
@mergify rebase

Mic92:
@mergify queue

PR #465, opened 2024-07-15
feat(zfs): set default networking.hostId

zimbatm:
In server scenario, disks are not getting shared, this is just not a thing. Don't force users to set this value.

### DIFF
--- a/nixos/common/zfs.nix
+++ b/nixos/common/zfs.nix
@@ -3,6 +3,9 @@
   # You may also find this setting useful to automatically set the latest compatible kernel:
   #boot.kernelPackages = config.boot.zfs.package.latestCompatibleLinuxPackages;
 
+  # Use the same default hostID as the NixOS install ISO and nixos-anywhere.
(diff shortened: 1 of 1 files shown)


nixos/common/zfs.nix
```diff
@@ -3,6 +3,9 @@
   # You may also find this setting useful to automatically set the latest compatible kernel:
   #boot.kernelPackages = config.boot.zfs.package.latestCompatibleLinuxPackages;
 
+  # Use the same default hostID as the NixOS install ISO and nixos-anywhere.
```
Mic92, line 6:
```suggestion
  # Use the same default hostID as the NixOS install ISO and nixos-anywhere.
  # This allows us to import zfs pool without using a force import.
  # ZFS has this as a safety mechanism for networked block storage (ISCSI), but 
  # in practice we found it causes more breakages like unbootable machines,
  # while people using ZFS on ISCSI is quite rare.
```

PR #466, opened 2024-07-16
break(nix): disable nix channels

zimbatm:
Channels are not reproducible and should be removed.

This change:
* removes the `nix-channel` binary
* sets an empty NIX_PATH
* stop adding the /root/.nix-channels file

This will lead scripts depending on channels to break, exposing their dependency on channels.

### DIFF
--- a/nixos/common/nix.nix
+++ b/nixos/common/nix.nix
@@ -1,7 +1,10 @@
 { lib, config, ... }:
 {
+  # Disable nix channels. Use flakes instead.
+  nix.channel.enable = lib.mkDefault false;
+
   # Fallback quickly if substituters are not available.
-  nix.settings.connect-timeout = 5;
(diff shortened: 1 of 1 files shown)


SuperSandro2000:
> * sets an empty NIX_PATH

that unfortunately currently breaks things https://github.com/NixOS/nixpkgs/pull/323613

Mic92:
With the pull request closed, is this now merge-able? https://github.com/NixOS/nixpkgs/pull/323613

phaer:
I believe it is! That PR was also backported, so we should be good with stable support.

One think we could consider in another PR is to set NIX_PATH and flake registry.json entries for config.srvos.flake and config.srvos.flake.inputs.nixpkgs where config.srvos.flake is set.

Issue #473, opened 2024-07-31
github DynamicUser cannot access nix-daemon


siriobalmelli:
Running the following workflow file: https://github.com/siriobalmelli-foss/ci-test/blob/main/.github/workflows/aarch64-nix.yml

With the following role config:

```nix
  roles.github-actions-runner = {
    url = "https://github.com/siriobalmelli-foss";
    count = 1;
    name = "ghrunner";
    githubApp = {
      id = "<ELIDED>";
      login = "<ELIDED>";
      privateKeyFile = config.sops.secrets.github.path;
    };
    cachix.cacheName = "siriobalmelli-nixpkgs";
    cachix.tokenFile = config.sops.secrets.cachix.path;
  };
```

Gives the following error:

```
user 'ghrunner-1' is not allowed to connect to the Nix daemon
```

I can work around this by adding:

```nix
nix.settings.allowed-users = ["ghrunner-1"];
```

I'm pretty sure the DynamicUser being created *should* be able to access `nix-daemon`, correct?

I'm unsure however whether I would try to rework https://github.com/nix-community/srvos/blob/1f867a5658bfc4318ea6f83304b2a1bc4a0b28ee/nixos/modules/github-runners/default.nix#L58 so that it adds an entry to `nix.settings.allowed-users` or whether I should take another approach; any suggestions welcome.

For context, this seems related to https://github.com/nix-community/srvos/issues/50

SuperSandro2000:
> I'm pretty sure the DynamicUser being created _should_ be able to access `nix-daemon`, correct?

That depends on the start order of the services. I think nix-daemon somehow only recognizes users that existed before it started. Generally I just remove all DynamicUsers that want to talk to the nix-daemon because of hard and seemingly time wasting bugs like this.

siriobalmelli:
@SuperSandro2000 I see, thank you.

@zimbatm from looking at https://github.com/nix-community/srvos/issues/50 it seems DynamicUser is being used for security reasons.

The use case I'm trying to support is GitHub runners which can run (and cache locally in their nix store) eg `nix flake check`, on aarch64 and x86_64.

Any suggestions what the best way forward would be? Once I better understand the correct approach I'm happy to file a PR.

zimbatm:
What supersandro suggested. Or add a dependency to the nix-daemon.service on this one to ensure the loading order.

Most of our work is now going into https://github.com/nix-community/buildbot-nix, so this module will probably need to be removed at some point.

PR #477, opened 2024-08-06
modules/github-runners: encode nix-daemon's dependency on runners

siriobalmelli:
Runners should be able to access nix-daemon so as to build using nix.

Services started after nix-daemon cannot access the nix store; runners use dynamic users which makes 'nix.settings.allowed-users' prone to failures.

See https://github.com/nix-community/srvos/issues/473

### DIFF
--- a/nixos/modules/github-runners/default.nix
+++ b/nixos/modules/github-runners/default.nix
@@ -73,5 +73,6 @@ in
         )
       )
     );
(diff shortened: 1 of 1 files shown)


siriobalmelli:
Withdrawing PR: dropped srvos entirely and using nixpkgs [github-runner](https://github.com/NixOS/nixpkgs/tree/master/nixos/modules/services/continuous-integration/github-runner) directly.

PR #478, opened 2024-08-06
chore(flake): update mkdocs-numtide

zimbatm:
Unfollow nixos-stable since mkdocs-numtide depends on a plugin that is only available on nixos-unstable.


zowoq:
https://github.com/nix-community/srvos/actions/runs/10274617784/job/28431883255

The docs should be cached for deployment after going through CI but this seems to have broken that somehow.

I've removed max-jobs so deployment works again.

PR #479, opened 2024-08-07
feat(docs): Add links to module docs

### LINKED ISSUE #476: docs request: linking directly to the modules
Hey y'all - thanks for putting this together. It's been great having a good reference for proper defaults. While I was browsing the docs, I saw that there are references to the code, like so: 

image

Unfortunately, this means that I also need to have a separate tab open to find this particular file and look through it. I see with `mkdocs` you have a variable for the `repo_url`. I attempted to template that in, but it seems like your version of mkdocs doesn't support this feature of Jinja templating, eg. `{{ repo_url }}/blob/main/nixos/hardware/hetzner-cloud/default.nix`. There are plugins for this ([like this one](https://github.com/rosscdh/mkdocs-markdownextradata-plugin)), but I don't know your appetite for going that far. :-)

What are your thoughts on just adding hardcoded links directly to modules in the Github repo? So the above: `nixosModules.hardware-hetzner-cloud` becomes: [`nixosModules.hardware-hetzner-cloud`](https://github.com/nix-community/srvos/blob/main/nixos/hardware/hetzner-cloud/default.nix)? I'm happy to put together a PR for it if so.

aos:
Closes #476

Here's a first pass at linking to the modules. I thought about finding a way to automate this (eg. populating the links directly from the exposed flake modules) but couldn't find a clean solution.

### DIFF
--- a/docs/darwin/getting_started.md
+++ b/docs/darwin/getting_started.md
@@ -4,8 +4,8 @@
 
 This project exports four big categories of NixOS modules which are useful to define a server configuration:
 
-* Machine type - these are high-level settings that define the machine type (Eg: common, server or desktop). Only one of those would be included.
-* Configuration mixins - these define addons to be added to the machine configuration. One or more can be added.
+* [Machine type](./type.md) - these are high-level settings that define the machine type (Eg: common, server or desktop). Only one of those would be included.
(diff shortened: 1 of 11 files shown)


zimbatm (approved):
Wonderful. Thank you

PR #481, opened 2024-08-08
initrd.systemd: disable emergency mode

mrVanDalo:
we disable emergency mode in systemd, but if systemd is enabled during boot we still end up in emergency mode eventually, this will fix that.

### DIFF
--- a/nixos/server/default.nix
+++ b/nixos/server/default.nix
@@ -1,6 +1,7 @@
 # A default configuration that applies to all servers.
 # Common configuration across *all* the machines
 {
+  config,
   pkgs,
   lib,
   options,
--- a/nixos/server/default.nix
+++ b/nixos/server/default.nix
@@ -73,6 +73,14 @@
   # No mutable users by default
   users.mutableUsers = false;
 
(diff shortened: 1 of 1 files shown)


nixos/server/default.nix
```diff
@@ -62,6 +62,14 @@
   # No mutable users by default
   users.mutableUsers = false;
 
+  # Given that our systems are headless, emergency mode is useless.
+  # We prefer the system to attempt to continue booting so
+  # that we can hopefully still access it remotely.
+  boot.initrd.systemd.suppressedUnits = [
```
Mic92, line 68:
```suggestion
  boot.initrd.systemd.suppressedUnits = lib.mkIf config.systemd.enableEmergencyMode [
```

Mic92: Users might still enable emergency mode in their configuration.

mrVanDalo:
I cleaned up a bit this commits and rebase to **main**

mrVanDalo:
Ok now everything is green.

zimbatm:
thanks!

PR #484, opened 2024-08-15
fix: set lower prio on networking.hostName

zimbatm:
Avoid clashes with the clan-core project.

1337 is in between mkDefault (1000) and mkOptionDefault (1500).

### DIFF
--- a/nixos/mixins/cloud-init.nix
+++ b/nixos/mixins/cloud-init.nix
@@ -26,5 +26,5 @@
   networking.useDHCP = lib.mkDefault false;
 
   # Delegate the hostname setting to cloud-init by default
-  networking.hostName = lib.mkOverride 1337 "";
+  networking.hostName = lib.mkOverride 1337 ""; # lower prio than lib.mkDefault
 }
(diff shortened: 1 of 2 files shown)


nixos/mixins/cloud-init.nix
```diff
@@ -26,5 +26,5 @@
   networking.useDHCP = lib.mkDefault false;
 
   # Delegate the hostname setting to cloud-init by default
-  networking.hostName = lib.mkDefault "";
+  networking.hostName = lib.mkOverride 1337 "";
```
Mic92, line 29:
```suggestion
  networking.hostName = lib.mkOverride 1337 ""; # lower prior lib.mkDefault
```

nixos/server/default.nix
```diff
@@ -43,7 +43,7 @@
   networking.firewall.enable = true;
 
   # Delegate the hostname setting to dhcp/cloud-init by default
-  networking.hostName = lib.mkDefault "";
+  networking.hostName = lib.mkOverride 1337 "";
```
Mic92, line 46:
```suggestion
  networking.hostName = lib.mkOverride 1337 ""; # lower prior lib.mkDefault
```

PR #485, opened 2024-08-16
nix-experimental: enable ca-derivations and impure-derivations

Mic92:


zowoq:
Done in https://github.com/nix-community/srvos/pull/631.

PR #486, opened 2024-08-17
nix: don't set nix.optimize.automatic in containers

Mic92:


Mic92:
Fixes:

 error: The option `nix.optimise.automatic' has conflicting definition values:
- In `/nix/store/66sb7685qiy21zzcjhq0c0z6y7wfkngh-source/nixos/common/nix.nix': true
- In `/nix/store/n1vpnmm4d56jmkx7x9dciqrd5rhq2w1l-source/nixos/modules/virtualisation/container-config.nix': false
Use `lib.mkForce value` or `lib.mkDefault value` to change the priority on any of these definitions.

zimbatm (approved):

PR #490, opened 2024-08-22
fix(nix-remote-builder): allow more connections

zimbatm:
Given the nix-daemon forking model, it can starve the remote builder of connections. Bump the limits.

See also https://github.com/nix-community/infra/pull/1417

### DIFF
--- a/nixos/roles/nix-remote-builder.nix
+++ b/nixos/roles/nix-remote-builder.nix
@@ -42,5 +42,8 @@ in
     users.users.nix-remote-builder.isNormalUser = true;
     users.users.nix-remote-builder.group = "nogroup";
     nix.settings.trusted-users = [ "nix-remote-builder" ];
+
+    # Allow more nix-daemon sessions to connect at the same time.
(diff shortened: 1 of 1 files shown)


Mic92 (approved):

phaer (commented):
Sounds like a good default for remote-builders to me, Thanks!

iiuc, `MaxStartups` is about connections, maybe we could also raise `MaxSessions = 100` for connections using ssh multiplexing (`ControlMaster yes`)?

Going too high with `MaxStartups` could maybe DoS attempts worse, as this is for unauthenticated sessions. If we assume pubkey authentication only, we could add `MaxAuthTries 1` (default 6) while at it.

Mic92:
Does nix uses ControlMaster or is compatible with it?

phaer:
It does with ssh-ng stores (at least my current Lix does, can check nix later)

zimbatm:
This needs another pass. The MaxStartups syntax is "start:rate:full" in the manual instead of a single number.

PR #491, opened 2024-08-25
{hetzner-cloud, vultr}: don't conflict with disko

Mic92:
boot.loader.grub.device maps internally to boot.loader.grub.devices. Disko only sets boot.loader.grub.devices. By using the default priority we allow disko to override this value.

fixes https://github.com/nix-community/disko/issues/748

### DIFF
--- a/nixos/hardware/hetzner-cloud/default.nix
+++ b/nixos/hardware/hetzner-cloud/default.nix
@@ -13,7 +13,7 @@
 
   config = {
     boot.growPartition = true;
-    boot.loader.grub.device = "/dev/sda";
+    boot.loader.grub.devices = lib.mkDefault [ "/dev/sda" ];
 
     fileSystems."/" = lib.mkDefault {
       device = "/dev/sda1";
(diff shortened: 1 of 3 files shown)


Mic92:
Tested in https://github.com/NixOS/nixos-wiki-infra/pull/134

nixos/hardware/hetzner-cloud/default.nix
```diff
@@ -13,7 +13,7 @@
 
   config = {
     boot.growPartition = true;
-    boot.loader.grub.device = "/dev/sda";
+    boot.loader.grub.devices = lib.mkDefault [ "/dev/sda" ];
```
zimbatm, line 16: This is a noop change because the grub module overrides this default:
https://github.com/NixOS/nixpkgs/blob/dd6c0ee128ae96cbb32855741ec85576a1368c1c/nixos/modules/system/boot/loader/grub/grub.nix#L715

That means that it works when Disko is being used but not otherwise.

I'm not sure what the best approach is to fix this. Make change the NixOS module?

Mic92: Should be fixed upstream to use lib.mkIf

Mic92: I would love to deprecate the singular version of grub.devices... It already has caused so much headache for me.

PR #495, opened 2024-08-29
remove mddoc

zowoq:
has been the default for a while

### DIFF
--- a/nixos/roles/prometheus/default.nix
+++ b/nixos/roles/prometheus/default.nix
@@ -38,7 +38,7 @@ in
                 type = lib.types.str;
                 default = name;
               };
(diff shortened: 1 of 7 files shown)


nixos/roles/prometheus/default.nix
```diff
@@ -83,7 +83,7 @@ in
                           type = lib.types.str;
                           default = name;
                         };
-                        enable = lib.mkEnableOption (lib.mdDoc "Enable recording rule") // {
+                        enable = lib.mkEnableOption ("Enable recording rule") // {
```
SuperSandro2000, line 86: The () here can now be dropped

PR #496, opened 2024-08-30
feat(srvos.flake): export the flake further

zimbatm:
When `srvos.flake` is set, also set:
* NIX_PATH to the flake's nixpkgs input
* flake-registry to the flake's inputs

As suggested in #466 by @phaer

### DIFF
--- a/nixos/common/flake.nix
+++ b/nixos/common/flake.nix
@@ -7,6 +7,21 @@
 
 let
   cfg = config.srvos;
+
+  flake-registry = pkgs.writers.writeJSON "flake-registry.json" {
+    version = 2;
+    flakes = pkgs.lib.mapAttrsToList (name: value: {
+      exact = true;
+      from = {
+        id = name;
+        type = "indirect";
+      };
+      to = {
+        path = toString value;
+        type = "path";
+      };
+    }) ({ self = cfg.flake; } // cfg.flake.inputs);
+  };
 in
 {
   options.srvos = {
@@ -42,6 +57,9 @@ in
     };
   };
   config = lib.mkIf (cfg.flake != null) {
+    nix.settings.nix-path = lib.mkIf (cfg.flake.inputs ? nixpkgs) "nixpkgs=${cfg.flake.inputs.nixpkgs}";
+    nix.settings.flake-registry = lib.mkDefault flake-registry;
+
     system.extraSystemBuilderCmds = lib.optionalString cfg.symlinkFlake ''
       ln -s ${cfg.flake} $out/flake
     '';


nixos/common/flake.nix
```diff
@@ -7,6 +7,21 @@
...
+      exact = true;
+      from = {
+        id = name;
+        type = "indirect";
+      };
+      to = {
+        path = toString value;
+        type = "path";
+      };
+    }) cfg.flake.inputs;
```
phaer, line 23:
```suggestion
    }) ({ self = cfg.flake; } // cfg.flake.inputs);
```

Or something like this? Not sure about the name, maybe `origin`, `system` or just `flake` would be a better choice? 

But I think it's useful to have the flake itself if available, i.e. to explore the systems config in a nix repl. Not sure whether we would even need the inputs, except nixpkgs, separately anymore if we do this.

nixos/common/flake.nix
```diff
@@ -42,6 +57,9 @@ in
     };
   };
   config = lib.mkIf (cfg.flake != null) {
+    nix.settings.nix-path = cfg.flake.inputs.nixpkgs;
```
phaer, line 60: This syntax works as intended? Just reviewing the code atm, haven't tested but would have expected something like
`nix.settings.nix-path = "nixpkgs=${cfg.flake.inputs.nixpkgs}";`

nixos/common/flake.nix
zimbatm: Sounds good. `self` also maps what you get in the output function.

nixos/common/flake.nix
zimbatm: Good call.

nixos/common/flake.nix
```diff
@@ -42,6 +57,9 @@ in
     };
   };
   config = lib.mkIf (cfg.flake != null) {
+    nix.settings.nix-path = lib.mkIf (cfg.flake.inputs ? nixpkgs) "nixpkgs=${cfg.flake.inputs.nixpkgs}";
```
Mic92, line 60: This can be a bit wonky in nix versions < 2.24: https://github.com/NixOS/nix/issues/9574#issue-2033997689

nixos/common/flake.nix
```diff
@@ -42,6 +57,9 @@ in
...
+    nix.settings.flake-registry = lib.mkDefault flake-registry;
```
Mic92, line 61: Not sure we want this by default because this will upload nixpkgs to all kinds of machines, which can overhead and slow to upload. There is also a proper nixos option for this: https://github.com/Mic92/dotfiles/blob/9e2f4683cdf6a942e8ffc65ed3aa5ef27aaa4652/nixos/modules/pinned-registry.nix#L3

Mic92: Just now the module was using builtins.unsafeDiscardStringContext

phaer:
Should probably iterate a bit more on this. Besides @Mic92 s comments, I also learned that there's options for this since 24.05 in https://github.com/NixOS/nixpkgs/blob/nixos-24.05/nixos/modules/misc/nixpkgs-flake.nix

nixos/common/flake.nix
Mic92: Also potentially Nixos has something like this now as well?

Mic92:
Ah. We should just switch to that.

phaer:
https://github.com/nix-community/srvos/pull/500

PR #500, opened 2024-09-05
break(flake): use upstream options...

phaer:
* use nixpkgs.flake settings to add nixpkgs
* drop symlink flake
* replace it with registerSelf, which uses the flake registry
* don't add all inputs of `self` anymore, they can quickly be referenced with `self.inputs` anyway.

### DIFF
--- a/nixos/common/flake.nix
+++ b/nixos/common/flake.nix
@@ -7,21 +7,6 @@
 
 let
   cfg = config.srvos;
-
-  flake-registry = pkgs.writers.writeJSON "flake-registry.json" {
-    version = 2;
-    flakes = pkgs.lib.mapAttrsToList (name: value: {
-      exact = true;
-      from = {
-        id = name;
-        type = "indirect";
-      };
-      to = {
-        path = toString value;
-        type = "path";
-      };
-    }) ({ self = cfg.flake; } // cfg.flake.inputs);
-  };
 in
 {
   options.srvos = {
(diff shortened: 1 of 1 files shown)


phaer:
I did no lib.mkRenamedOptionModule for symlinkFlake -> registerSelf, as it's not just a rename.  

We say "stable" in the README, but we don't really have releases or anything to communicate 
breaking changes yet, I think? Technically, there's another breaking change as this doesn't add all inputs of `self` anymore - on the main branch for a few days.

zimbatm:
Ok. I assumed because you were duplicating some of the stuff in https://github.com/nix-community/infra/pull/1436

zimbatm (approved):

nixos/common/flake.nix
```diff
@@ -55,14 +40,20 @@ in
...
-    nix.settings.flake-registry = lib.mkDefault flake-registry;
 
-    system.extraSystemBuilderCmds = lib.optionalString cfg.symlinkFlake ''
-      ln -s ${cfg.flake} $out/flake
-    '';
+    nixpkgs.flake = {
+      source = lib.mkDefault (cfg.flake.inputs.nixpkgs or null);
+    };
+
+    nix.registry = lib.optionalAttrs cfg.registerSelf {
```
Mic92, line 51: I think we can actually enable this for the desktop profile?

PR #502, opened 2024-09-07
Add assertion for custom sudo rules

DavHau:
I added custom rules to by sudo to allow a single user to restart a service.

Since srvos enables `security.sudo.execWheelOnly` by default, this breaks such use cases and it took me a while to debug it until I found out about this option.

Basically any sudo rule declared on anything else than `root` or `wheel` will not have any effect until this option is disabled again.

I now added an assertion to check for this scenario and crash with a nice error telling you to disable `execWheelOnly` if you declared any such rules


SuperSandro2000:
Is that something that could be upstreamed into nixpkgs?

Mic92:
Yes.

PR #507, opened 2024-09-14
run nix-gc with low io/cpu priority

Mic92:


zimbatm (approved):

zimbatm:
this could be upstreamed

Mic92:
I like to test the impact of these options in smaller deployments first before upstreaming.

PR #509, opened 2024-09-15
no longer set networking.useDHCP to false

Mic92:
This was originally intended to disable scripted networking, but it's actually also an option that systemd-networkd respects. So we shouldn't have it.


Mic92:
@zimbatm was the original idea to disable this to not use dhcpcd?

zimbatm:
As I remember, useDHCP used to enable the old subsystem. It should be removed if it disables it in networkd.

PR #514, opened 2024-09-19
Make sure the serial console is visible in qemu when testing the server

pingiun:
I found this very useful in my testing of server configurations

### DIFF
--- a/nixos/server/default.nix
+++ b/nixos/server/default.nix
@@ -125,5 +125,5 @@
 
   # Make sure the serial console is visible in qemu when testing the server configuration
   # with nixos-rebuild build-vm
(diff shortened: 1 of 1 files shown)


nixos/server/default.nix
```diff
@@ -122,4 +122,8 @@
     "net.core.default_qdisc" = "fq";
     "net.ipv4.tcp_congestion_control" = "bbr";
   };
+
+  # Make sure the serial console is visible in qemu when testing the server configuration
+  # with nixos-rebuild build-vm
+  virtualisation.vmVariant.virtualisation.graphics = false;
```
Mic92, line 128:
```suggestion
  virtualisation.vmVariant.virtualisation.graphics = lib.mkDefault false;
```

PR #523, opened 2024-09-26
decrease watchdog runtimeTime for Raspberry Pi compatibility

danjujan:
Lower the watchdog interval to be Raspberry Pi compatible. See https://pimylifeup.com/raspberry-pi-watchdog/

### DIFF
--- a/nixos/server/default.nix
+++ b/nixos/server/default.nix
@@ -100,10 +100,10 @@
     #   https://0pointer.de/blog/projects/watchdog.html
     watchdog = {
       # systemd will send a signal to the hardware watchdog at half
-      # the interval defined here, so every 10s.
-      # If the hardware watchdog does not get a signal for 20s,
+      # the interval defined here, so every 7.5s.
+      # If the hardware watchdog does not get a signal for 15s,
       # it will forcefully reboot the system.
-      runtimeTime = "20s";
+      runtimeTime = "15s";
       # Forcefully reboot if the final stage of the reboot
       # hangs without progress for more than 30s.
       # For more info, see:


SuperSandro2000:
Doesn't this make it more trigger happy for all other arches? Can we maybe detect PIs and only lower it for them?

sedlund:
maybe as a hardware module specific for Raspi in https://github.com/nix-community/srvos/tree/main/nixos/hardware ?

It should be documented as its a limitation of the bcm2835_wdt driver.  I believe the limit is actually 16 seconds.

https://github.com/torvalds/linux/blob/075dbe9f6e3c21596c5245826a4ee1f1c1676eb8/drivers/watchdog/bcm2835_wdt.c#L29

```c
#define PM_WDOG_TIME_SET		0x000fffff   (1048576)
#define WDOG_TICKS_TO_MSECS(x) ((x) * 1000 >> 16)

.max_hw_heartbeat_ms =	WDOG_TICKS_TO_MSECS(PM_WDOG_TIME_SET),
```
(1048576 * 1000) / 2^16 = 16000 ms
16000/1000 = 16sec

Also, I have had to disable watchdog on pi when doing nixos activations that have many systemd restarts with slow SD IO which causes the system to reset before the activation completes.

sedlund:
Rethinking this. If you set the timer longer than 16 seconds, it will be set to 16 seconds due to the limit imposed above.

This patch does nothing for pi.  Which is actually kind of nice.  💖

Mic92:
Can you make a raspberry hardware profile for this? https://github.com/nix-community/srvos/tree/main/nixos/hardware

Mic92:
Ok. Let's just set it to 15s instead...

Mic92:
I see now an increased failure rate on x86 based servers. Ipmi shows this is caused by watchdogs.

Mic92:
I will try to isolate this a bit

PR #528, opened 2024-09-29
enable fstrim by default

Mic92:


nixos/common/default.nix
```diff
@@ -20,6 +20,10 @@
     enableNg = lib.mkDefault true;
   };
 
+  # The majority of storage used for root filesystem is based on Sata SSD/NVMe,
+  # which benefits from enabling the periodic filesystem TRIM operation.
+  services.fstrim.enable = lib.mkDefault true;
```
zowoq, line 25: Is this needed on zfs with zpool trim?

Mic92: We have a similar service called zpool-trim.service that is enabled by default.
However since it's just run once a week, its basically free even if you don't need it, given the executable is also part of util-linux.

PR #539, opened 2024-10-03
add shared update-diff

zowoq:
also rename the files to match the option


zimbatm (approved):

zimbatm:
One thing I don't like here is that the shared code is not standalone. Now you need to know that to use the update-diff you also need to include the platform-specific code.

zowoq:
Sorry, I don't understand what you mean?

PR #560, opened 2024-11-18
chore(nix): make the project usable without flakes

zimbatm:
The issue with the current dev-flake setup is that `builtins.getFlake` is not available in Nix, unless flakes are enabled. So this idea of using flake-compat doesn't work.

Since the project is exporting modules, it doesn't matter. Replace the top-level default.nix flake-compat with a simple attrset.


flake.nix
```diff
@@ -22,100 +22,102 @@
...
+      modules = import ./.;
     in
-    inputs.flake-parts.lib.mkFlake { inherit inputs; } {
-      imports = [
-        inputs.git-hooks-nix.flakeModule
-        inputs.treefmt-nix.flakeModule
-      ];
+    # builtins.getFlake is not available in flake-compat mode
+    if builtins ? getFlake then
+      inputs.flake-parts.lib.mkFlake { inherit inputs; } {
```
Mic92, line 31: Actually, Can we remove inputs.flake-parts from here? Because it introduces a hidden nixpkgs that we cannot override. Here is an example how this can be separated: https://github.com/NixOS/nix/blob/3b76d01f3be01792dc45a89fc4266a605494ed84/flake.nix#L80

Mic92: Someone did already contact me about that not long ago.

zimbatm:
@Mic92 like this?

Mic92 (approved):

flokli:
The whole "dev flake" business is due to https://github.com/NixOS/nix/issues/6124, it'd be great to have a way to specify non-propagated dev-only flake inputs.

PR #567, opened 2024-11-24
zfs: drop 'latestCompatibleLinuxPackages'

jkachmar:
## description

exactly what it says in the commit message: this config field has been deprecated as of 24.11.

since it was only ever a commented out suggestion, it should be safe to remove even before `srvos` updates its nixpkgs snapshot.


zimbatm:
thanks

PR #573, opened 2024-12-02
update to 24.11

Mic92:


dev/default.nix
```diff
@@ -60,7 +60,7 @@
       };
       packages =
         {
-          update-dev-private-narHash = pkgs.writeScriptBin "update-dev-private-narHash" ''
+          update-dev-private-narHash = pkgs.writeShellScriptBin "update-dev-private-narHash" ''
```
Mic92, line 63: Since this depends on private flakes itself, it's a bit of a catch-22

Mic92:
@mergify queue

PR #575, opened 2024-12-02
userborn: enable by default

Mic92:
I have tested this for a wile without any issues.
If it works find in srvos, we can propose this to nixos itself.


nixos/common/default.nix
```diff
@@ -15,6 +20,10 @@
     ./zfs.nix
   ];
 
+  # Create users with https://github.com/nikstur/userborn rather than our perl script.
+  # Don't enable if we detect impermanence, which is not compatible with it: https://github.com/nix-community/impermanence/pull/223
+  services.userborn.enable = lib.mkIf (options.environment ? persistence) (lib.mkDefault true);
```
zowoq, line 25:
```suggestion
  services.userborn.enable = lib.mkIf (!options.environment ? persistence) (lib.mkDefault true);
```

zowoq:
Causes nixos-rebuild switch / nixos-rebuild-switch-to-configuration.service to fail?

I tried it on build01:

`switch` fails.

`boot` works but then running `switch` again (switching into the current booted generation) fails?

Mic92:
What is the failure?

zowoq:
```
Dec 05 00:42:37 build01 systemd[1]: Starting /nix/store/k88rz3p58mp584jd0v63wlxx3vja2fwg-nixos-system-build01-25.05.20241201.767b0e3/bin/switch-to-configuration switch...
Dec 05 00:42:37 build01 systemd[1]: Started /nix/store/k88rz3p58mp584jd0v63wlxx3vja2fwg-nixos-system-build01-25.05.20241201.767b0e3/bin/switch-to-configuration switch.
Dec 05 00:42:38 build01 nixos[4248]: switching to system configuration /nix/store/k88rz3p58mp584jd0v63wlxx3vja2fwg-nixos-system-build01-25.05.20241201.767b0e3
Dec 05 00:42:40 build01 nixos[4248]: switching to system configuration /nix/store/k88rz3p58mp584jd0v63wlxx3vja2fwg-nixos-system-build01-25.05.20241201.767b0e3 failed (status 2)
Dec 05 00:42:40 build01 systemd[1]: nixos-rebuild-switch-to-configuration.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Dec 05 00:42:40 build01 systemd[1]: nixos-rebuild-switch-to-configuration.service: Failed with result 'exit-code'.
```

Mic92:
@zowoq are you sure this is caused by userborn?

zowoq:
Yes.

Mic92:
The error message doesn't give anything away. I would have expected all userborn runs to also have an error message.

Mic92:
The only exit code here is 1: https://github.com/nikstur/userborn/blob/2c4dca6cac5e974727e2b6b22bab2fd2ebe991f3/rust/userborn/src/main.rs#L55
And it should always print an error. The action script exits with 2 however.

zowoq:
I tried enabling it on a personal system and build04, same error on both.

Mic92:
Ok. This is reason enough to not enable it by default for now.

zowoq (approved):

nixos/common/default.nix
```diff
@@ -15,6 +20,13 @@
     ./zfs.nix
   ];
 
+  # Create users with https://github.com/nikstur/userborn rather than our perl script.
+  # Don't enable if we detect impermanence, which is not compatible with it: https://github.com/nix-community/impermanence/pull/223
+  # as well as agenix: https://github.com/ryantm/agenix/pull/255
+  services.userborn.enable = lib.mkIf (!(options.environment ? persistence || options ? age)) (
```
SuperSandro2000, line 26: Shouldn't we also check for environment.persistence.enable and

Mic92:
@mergify queue

Mic92:
@mergify queue

SuperSandro2000 (approved):
For my config this just worked but it still feels a bit "alpha" or at least beta

Mic92:
> For my config this just worked but it still feels a bit "alpha" or at least beta

How so? I am using this since September without any issues?

SuperSandro2000:
Found a small issue https://github.com/nikstur/userborn/pull/14 but maybe I just forget about it in the next weeks and it really just works.

phaer:
I believe this should be good to merge? Or are there any objections left?

SuperSandro2000:
We want to merge the linked PR in https://github.com/nikstur/userborn/issues/13#issuecomment-2536216618 and do something about https://github.com/nikstur/userborn/issues/7#issuecomment-2684368337 and then push for a release. I've been using this for a while with no extra problems.

Mic92:
> We want to merge the linked PR in [nikstur/userborn#13 (comment)](https://github.com/nikstur/userborn/issues/13#issuecomment-2536216618) and do something about [nikstur/userborn#7 (comment)](https://github.com/nikstur/userborn/issues/7#issuecomment-2684368337) and then push for a release. I've been using this for a while with no extra problems.

We already agenix, so this is just about subuids.

nikstur:
Yes I agree and understand that something needs to be done to fix the SUBUIDs/SUBGIDs issue and I see this within the responsibility of Userborn. I just want to understand whether this is needed for srvos, i.e. whether this is a blocker.

aanderse:
fwiw `userborn` rocks and has been nothing but great for me... thanks again ❤️

Mic92:
> Yes I agree and understand that something needs to be done to fix the SUBUIDs/SUBGIDs issue and I see this within the responsibility of Userborn. I just want to understand whether this is needed for srvos, i.e. whether this is a blocker.

I think we just need to not enable userborn if any of the options that userborn doesn't support is enabled.

nixos/common/default.nix
```diff
@@ -16,6 +21,16 @@
     ./zfs.nix
   ];
 
+  # Create users with https://github.com/nikstur/userborn rather than our perl script.
+  # Don't enable if we detect impermanence, which is not compatible with it: https://github.com/nix-community/impermanence/pull/223
```
nikstur: I don't understand why Userborn would be incompatible with impermanence. You can just point `passwordFilesLocation` to your persistent partition.

phaer: Not an impermanence user myself, so I don't really have a motivation to read the linked PR in detail. But one argument to keep it disabled with impermanence, even if the comment isn't entirely accurate, would be that you seem to say it requires a custom option. And as this PR is about a default we should assume default settings for other modules as well as far as feasible. Users are still free to opt into userborn. (as many of us already do)

PR #577, opened 2024-12-02
 chore: move update-private-narhash to script 

zimbatm:
The issue with the current dev-flake setup is that `builtins.getFlake`
is not available in Nix, unless flakes are enabled. So this idea
of using flake-compat doesn't work.

Since the project is exporting modules, it doesn't matter. Replace the
top-level default.nix flake-compat with a simple attrset.

### DIFF
--- a/dev/update-private-narhash
+++ b/dev/update-private-narhash
@@ -6,7 +6,7 @@ cd "$(dirname "$0")"
 
 echo "Updating $PWD/private.narHash" >&2
 
-nix flake lock ./private
-nix hash path ./private > ./private.narHash
+nix --extra-experimental-flags 'flakes nix-command' flake lock ./private
+nix --extra-experimental-flags 'flakes nix-command' hash path ./private > ./private.narHash
 
 echo OK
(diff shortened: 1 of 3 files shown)


dev/update-private-narhash
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+# Used to update the private dev flake hash reference.
+set -euo pipefail
+
+cd "$(dirname "$0")"
+
+echo "Updating $PWD/private.narHash" >&2
+
+nix flake lock ./private
```
Mic92, line 9:
```suggestion
nix --extra-experimental-flags 'flakes nix-command' flake lock ./private
```

dev/update-private-narhash
```diff
@@ -0,0 +1,12 @@
...
+nix hash path ./private > ./private.narHash
```
Mic92, line 10:
```suggestion
nix --extra-experimental-flags 'flakes nix-command' hash path ./private > ./private.narHash
```

Mic92:
@mergify queue

dev/update-private-narhash
zowoq: features, not flags.

https://github.com/nix-community/srvos/commit/a6ac04bf8ac513d8e4a506d7d69c0f161e4b528c

PR #578, opened 2024-12-02
common/flake: include cfg.flake in telegraf inputs

phaer:
We currently only expose the flake's inputs, but not the flake itself which is useful to check the currently flake via monitoring.

### DIFF
--- a/shared/common/flake.nix
+++ b/shared/common/flake.nix
@@ -57,7 +57,9 @@ in
 
     services.telegraf.extraConfig.inputs.file =
       let
-        inputsWithDate = lib.filterAttrs (_: input: input ? lastModified) cfg.flake.inputs;
+        inputsWithDate = lib.filterAttrs (_: input: input ? lastModified) (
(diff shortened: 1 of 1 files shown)


shared/common/flake.nix
```diff
@@ -57,7 +57,9 @@ in
 
     services.telegraf.extraConfig.inputs.file =
       let
-        inputsWithDate = lib.filterAttrs (_: input: input ? lastModified) cfg.flake.inputs;
+        inputsWithDate = lib.filterAttrs (_: input: input ? lastModified) (
+          cfg.flake.inputs // { inherit (cfg) flake; }
```
Mic92, line 61: Doesn't this generate a rebuild on every commit? I would prefer not to having to rebuild all my machines everytime.

phaer: But it's only relevant if srvos.flake is set and in that case you already get rebuilds on each commit, as the flake is included in the closure, no?

Or are you referring to telegraf restarts specificially?

PR #587, opened 2024-12-25
depend on gitMinimal instead of git

Mic92:
git requires requires in graphics libraries dependencies etc.

### DIFF
--- a/nixos/server/default.nix
+++ b/nixos/server/default.nix
@@ -15,7 +15,7 @@
   ];
 
   # used in shared/server.nix
-  config.programs.git.package = lib.mkDefault pkgs.gitMinimal;
+  programs.git.package = lib.mkDefault pkgs.gitMinimal;
 
   environment = {
     # Print the URL instead on servers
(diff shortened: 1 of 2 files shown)


SuperSandro2000 (approved):

zimbatm:
We used to have something like that, but it led to having two copies of git since other programs include git in their runtime closure.

nixos/server/default.nix
```diff
@@ -14,6 +14,9 @@
     ../../shared/server.nix
   ];
 
+  # used in shared/server.nix
+  config.programs.git.package = lib.mkDefault pkgs.gitMinimal;
```
zimbatm, line 18:
```suggestion
  programs.git.package = lib.mkDefault pkgs.gitMinimal;
```

zimbatm (approved):
let's find out

Mic92:
> let's find out

For met it was the other way around. I had gitMinimal first in my closure. But in any case, users now can configure which git version they want.

PR #591, opened 2024-12-26
feat: add hostname change detection

zimbatm:
NixOS is just too easy to deploy to the wrong host. Eg:

    nixos-rebuild --flake .#wronghostname --target-host root@otherhost switch

It happened to me a few times, and when I started talking about it I
got the same feedback from other people.

After this change, the user now has to confirm with YES before the
switch can happen.

To disable, set

    srvos.detect-hostname-change.enable = false;

### DIFF
--- a/nixos/common/detect-hostname-change.nix
+++ b/nixos/common/detect-hostname-change.nix
@@ -24,6 +24,11 @@
         exit
       fi
 
+      # Useful for automation
+      if [[ "''${EXPECTED_HOSTNAME:-}" = "$desired" ]];
+        exit
+      fi
+
       log() {
         echo "$*" >&2
       }
--- a/nixos/common/detect-hostname-change.nix
+++ b/nixos/common/detect-hostname-change.nix
@@ -1,0 +1,43 @@
+# Protection against deploying system closures to the wrong host.
+{
+  config,
+  lib,
+  ...
+}:
+{
+  options.srvos.detect-hostname-change = {
+    enable = lib.mkEnableOption "warn if the hostname changes between deploys" // {
+      default = true;
+    };
+  };
+
+  config = lib.mkIf (config.srvos.detect-hostname-change.enable && config.networking.hostName != "") {
+    system.preSwitchChecks.detectHostnameChange = ''
+      # Ignore if the system is getting installed
+      if [[ ! -e /run/current-system ]]; then
+        exit
+      fi
+
+      actual=$(< /proc/sys/kernel/hostname)
+      desired=${config.networking.hostName}
+      if [[ "$actual" = "$desired" ]]; then
+        exit
+      fi
+
+      log() {
+        echo "$*" >&2
+      }
+
+      log "WARNING: machine hostname change detected from '$actual' to '$desired'"
+      log
+      log "Are you deploying on the right host?"
+      log
+      log "Type YES to continue:"
+      read reply
+      if [[ $reply != YES ]]; then
+        echo "aborting"
+        exit 1
+      fi
+    '';
+  };
+}
--- a/nixos/common/default.nix
+++ b/nixos/common/default.nix
@@ -5,6 +5,7 @@
 
   imports = [
     ../../shared/common/flake.nix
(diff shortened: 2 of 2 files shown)


nixos/common/detect-hostname-change.nix
```diff
@@ -0,0 +1,43 @@
...
+  options.srvos.detect-hostname-change = {
+    enable = lib.mkEnableOption "warn if the hostname changes between deploys" // {
+      default = true;
+    };
+  };
+
+  config = lib.mkIf (config.srvos.detect-hostname-change.enable && config.networking.hostName != "") {
+    system.preSwitchChecks.detectHostnameChange = ''
+      # Ignore if the system is getting installed
+      if [[ ! -e /run/current-system ]]; then
```
zowoq, line 17: This handles installing from non-nixos, what about installing from nixos, e.g. nixos-anywhere kexec?

zimbatm: Good question. Is there a way to detect the kexec environment?

phaer:
Thank you!

The warning plus confirmation is a nice idea, but it does add an requirement for interactive input (i.e. a tty) which isn't otherwise needed to switch in most configs?

Maybe an env variable like `NIXOS_NEW_HOSTNAME` or so might make sense as an escape hatch?


(TIL `system.preSwitchChecks`)

zimbatm:
How about `EXPECTED_HOSTNAME=foo` where "foo" is the target hostname?

nixos/common/detect-hostname-change.nix
```diff
@@ -0,0 +1,48 @@
...
+      fi
+
+      actual=$(< /proc/sys/kernel/hostname)
+      desired=${config.networking.hostName}
+      if [[ "$actual" = "$desired" ]]; then
+        exit
+      fi
+
+      # Useful for automation
+      if [[ "''${EXPECTED_HOSTNAME:-}" = "$desired" ]];
```
priegger, line 28: @zimbatm : This line is missing a "then", which makes bash exit with a parsing error 2 lines below.

I thought I wrote this comment somewhere on GitHub about 2 minutes ago, but can't find it anymore. Better ping twice than never I guess. ;-)

zowoq: https://github.com/nix-community/srvos/pull/595

zowoq:
> Maybe an env variable like NIXOS_NEW_HOSTNAME or so might make sense as an escape hatch?

> How about `EXPECTED_HOSTNAME=foo` where "foo" is the target hostname?

The `preSwitchChecks` step has a built in escape hatch so that could be used instead? https://github.com/NixOS/nixpkgs/commit/6e192c4489ea44c48876faa11060798fb9502146

phaer:
Ah yes, NIXOS_NO_CHECK would probably have been good enough. Did not know about that.

Kranzes:
Btw this broke nixos-anywhere.

```
setting up /etc...
WARNING: machine hostname change detected from 'nixos-installer' to 'qa'

Are you deploying on the right host?

aborting
Type YES to continue:
Pre-switch check 'detectHostnameChange' failed
Pre-switch checks faile
```
It didn't actually let me type in `YES`.

zimbatm:
An alternative design would be to source the info from `/etc/hostname`. That might be a better solution when using nixos-anywhere since the `/proc` filesystem is leaking the info from the host. The reason I didn't go for it is that I wasn't sure in what conditions the file existed. And also the /proc filesystem is bound to be more up to date.

Mic92:
> Btw this broke nixos-anywhere.
> 
> ```
> setting up /etc...
> WARNING: machine hostname change detected from 'nixos-installer' to 'qa'
> 
> Are you deploying on the right host?
> 
> aborting
> Type YES to continue:
> Pre-switch check 'detectHostnameChange' failed
> Pre-switch checks faile
> ```
> 
> It didn't actually let me type in `YES`.

Was a re-installation? Because I wonder why this exists:

```
if [[ ! -e /run/current-system ]]; then
```

PR #594, opened 2025-01-02
server: add option to re-enable documentation

zowoq:
Having a single option for the docs would be a bit simpler than needing to re-enable all the separate options.

I'll use this option to enable docs on the community builders.

### DIFF
--- a/nixos/common/default.nix
+++ b/nixos/common/default.nix
@@ -20,9 +20,6 @@
   # - for containers we currently rely on the `stage-2` init script that sets up our /etc
   boot.initrd.systemd.enable = lib.mkDefault (!config.boot.swraid.enable && !config.boot.isContainer);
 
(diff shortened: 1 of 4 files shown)


nixos/common/default.nix
```diff
@@ -20,9 +20,6 @@
   # - for containers we currently rely on the `stage-2` init script that sets up our /etc
   boot.initrd.systemd.enable = lib.mkDefault (!config.boot.swraid.enable && !config.boot.isContainer);
 
-  # Work around for https://github.com/NixOS/nixpkgs/issues/124215
-  documentation.info.enable = false;
```
zowoq, line 24: This work around is a couple of years old now so I'm assuming that it is safe to remove?

zimbatm (approved):

PR #596, opened 2025-01-08
nixos/update-diff: activationScripts -> preSwitchChecks

zowoq:
Using `preSwitchChecks` makes the diff work for switch-to-configuration `boot` as well as `switch`.

### DIFF
--- a/darwin/common/update-diff.nix
+++ b/darwin/common/update-diff.nix
@@ -5,8 +5,9 @@
   ];
 
   config = lib.mkIf config.srvos.update-diff.enable {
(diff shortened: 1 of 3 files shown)


phaer (approved):
Haven't checked, but I assume `preSwitchChecks` work with `--dry-run` as well?

Mic92 (approved):
Also haven't tested, but should be fine assuming that this comes from nix-community infra.

zowoq:
> Haven't checked, but I assume `preSwitchChecks` work with `--dry-run` as well?

Yes, `dry-activate` still works.

PR #600, opened 2025-01-16
prometheus: fix vmstat_pgmajfault

zowoq:


nixos/roles/prometheus/default-alerts.nix
```diff
@@ -212,7 +212,7 @@
       };
 
       HostMemoryUnderMemoryPressure = {
-        expr = "rate(node_vmstat_pgmajfault[1m]) > 1000";
```
Mic92, line 215: This one is node-exporter, so probably never worked.

PR #601, opened 2025-01-19
detect-hostname-change: improve installer detection

Enzime:
Currently the detection fails if we're trying to do an installation from a NixOS system or installer, the only time I can see that check working is if someone was trying to run `nixos-install` directly from a non-NixOS Linux distro which would also require they have Nix installed which would be super uncommon

### DIFF
--- a/nixos/common/detect-hostname-change.nix
+++ b/nixos/common/detect-hostname-change.nix
@@ -13,13 +13,16 @@
 
   config = lib.mkIf (config.srvos.detect-hostname-change.enable && config.networking.hostName != "") {
     system.preSwitchChecks.detectHostnameChange = ''
+      actual=$(< /proc/sys/kernel/hostname)
+
       # Ignore if the system is getting installed
-      if [[ ! -e /run/current-system ]]; then
+      # https://github.com/nix-community/nixos-images/blob/2fc023e024c0a5e8e98ae94363dbf2962da10886/nix/installer.nix#L12-L13
+      if [[ ! -e /run/booted-system || "$actual" == "nixos-installer" ]]; then
         exit
       fi
 
-      actual=$(< /proc/sys/kernel/hostname)
       desired=${config.networking.hostName}
+
       if [[ "$actual" = "$desired" ]]; then
         exit
       fi


zowoq:
The disko installTest uses `machine` as the initial hostname, should that be added here or add `NIXOS_NO_CHECK` in disko?

nixos/common/detect-hostname-change.nix
```diff
@@ -13,13 +13,16 @@
 
   config = lib.mkIf (config.srvos.detect-hostname-change.enable && config.networking.hostName != "") {
     system.preSwitchChecks.detectHostnameChange = ''
+      actual=$(< /proc/sys/kernel/hostname)
+
       # Ignore if the system is getting installed
-      if [[ ! -e /run/current-system ]]; then
+      # https://github.com/nix-community/nixos-images/blob/2fc023e024c0a5e8e98ae94363dbf2962da10886/nix/installer.nix#L12-L13
+      if [[ ! -e /run/current-system || "$actual" == "nixos" || "$actual" == "nixos-installer" ]]; then
```
phaer, line 20: Why are we including `nixos` here?

Enzime: By default the NixOS installer ISOs use `nixos` as the hostname

phaer: That's not specific to the installer, but just the default for networking.hostName if unset.

So if your goal here is to check for installers, it's probably preferable to check /etc/os-release for VARIANT_ID as nixos-anywhere does. Or an other method which isn't as likely to yield false matches.

If it's even needed, the check is easy to disable during install via NIXOS_NO_CHECK

I proposed that for nixos-anywhere in https://github.com/nix-community/nixos-anywhere/pull/447

Enzime: We can't check `/etc/os-release` because this runs as part of activation which runs inside the `nixos-enter` `chroot`

Instead after some testing I realized that `/run/booted-system` won't exist when installing as it gets created by stage 2 normally

I've also removed the `nixos` hostname check

PR #611, opened 2025-02-13
nginx: disable zstd

fpletz:
Should not be enabled by default because important fixes are not getting merged. Trying to figure out that zstd is the culprit is tedious. Enabling the zstd module breaks SSI (used by jitsi-meet) and home-assistant.

See also:
- https://github.com/NixOS/nixpkgs/issues/359412
- https://github.com/NixOS/nixpkgs/pull/381037
- https://github.com/NixOS/nixpkgs/pull/381678


zimbatm:
Makes sense, thanks @fpletz

PR #625, opened 2025-03-19
darwin/server: includeUninstaller -> darwin-uninstaller

zowoq:
https://github.com/LnL7/nix-darwin/commit/9951b44d5bdebd062dab70701d7b21350cbbc6ab

### DIFF
--- a/darwin/server/default.nix
+++ b/darwin/server/default.nix
@@ -15,7 +15,7 @@
   programs.man.enable = lib.mkDefault config.srvos.server.docs.enable;
 
   # remove uninstaller, use 'nix run github:LnL7/nix-darwin#darwin-uninstaller'
-  system.includeUninstaller = lib.mkDefault false;
+  system.tools.darwin-uninstaller.enable = lib.mkDefault false;
 
   # If the user is in @admin they are trusted by default.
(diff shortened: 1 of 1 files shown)


darwin/server/default.nix
```diff
@@ -15,7 +15,7 @@
   programs.man.enable = lib.mkDefault config.srvos.server.docs.enable;
 
   # remove uninstaller, use 'nix run github:LnL7/nix-darwin#darwin-uninstaller'
-  system.includeUninstaller = lib.mkDefault false;
+  system.tools.darwin-uninstaller.enable = lib.mkDefault false;
```
Enzime, line 18: Out of curiosity, is there a reason in particular you don't want to include the uninstaller by default? Part of the reason we include it by default is so that you don't need a working Nix that can build derivations in case a user accidentally uninstalled Nix before nix-darwin

Enzime: See https://github.com/LnL7/nix-darwin/issues/804 for more context

zowoq: The `server` modules disable stuff like this, docs, etc to improve closure size and eval.

PR #630, opened 2025-04-16
fix: sudoers are trusted in desktops too

yajo:
I was surprised to see that a sudoer in a desktop environment is not trusted by default. This sane setting applies in all environments, including desktops.

@moduon MT-9339

### DIFF
--- a/darwin/common/nix.nix
+++ b/darwin/common/nix.nix
@@ -11,4 +11,7 @@
       Minute = 45;
     }
   ];
+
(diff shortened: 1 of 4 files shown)


zowoq:
https://github.com/nix-community/srvos/blob/c6808c5b6575ae3343dc3c21277080f68a574bd6/darwin/server/default.nix#L20-L21

Please also move the trusted-user darwin/server to darwin/common/nix.

yajo:
Done.

PR #654, opened 2025-06-23
add description for update-diff.text

sedlund:
so {option}`documentation.nixos.includeAllModules` doesn't complain

### DIFF
--- a/shared/common/update-diff.nix
+++ b/shared/common/update-diff.nix
@@ -12,6 +12,7 @@
     };
     text = lib.mkOption {
       type = lib.types.str;
+      description = "diff script snippet";
     };
   };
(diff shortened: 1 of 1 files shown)


shared/common/update-diff.nix
```diff
@@ -12,6 +12,7 @@
     };
     text = lib.mkOption {
       type = lib.types.str;
+      description = "diff script snippet that runs pre-switch via {option}`system.preSwitchChecks`";
```
zowoq, line 15:
```suggestion
      description = "diff script snippet";
```
This option applies to darwin and nixos but pre-switch check is nixos only, it uses an activation script on darwin.

PR #696, opened 2025-09-18
build ghostty terminfo from bin package

Mic92:
Doens't require special deps


phaer:
Looks like we need to feature-gate this for -unstable or backport ghostty-bin to -stable?

zimbatm:
ghostty-bin.terminfo could be upstreamed as well

PR #700, opened 2025-09-30
perf(swap): make desktops less swappy and more snappy

yajo:
As [the Ubuntu docs][1] explain, 10 is the recommended swappiness for a
desktop computer. The system will look snappier when recovering from
situations where it had to swap.

[1]:
https://help.ubuntu.com/community/SwapFaq#What_is_swappiness_and_how_do_I_change_it.3F

### DIFF
--- a/nixos/desktop/default.nix
+++ b/nixos/desktop/default.nix
@@ -11,4 +11,6 @@
 
   # improve desktop responsiveness when updating the system
   nix.daemonCPUSchedPolicy = "idle";
+
+  boot.kernel.sysctl."vm.swappiness" = lib.mkDefault 10; # Default is 60, lower is less eager to swap
 }


Momi-V:
The linked documentation actively contradicts the official and more technically detailed documentation on [kernel.org](https://docs.kernel.org/admin-guide/sysctl/vm.html#swappiness). It also hasn't been substantially updated since 2008, when RAM was like 2GB, Swap was on HDDs, and zSwap wasn't an option.

It's probably a much better idea to make a PR to enable zSwap **(not zRam)** by default instead of playing with `vm.swappiness`

Ubuntu:
> swappiness can have a value of between 0 and 100

Kernel:
> This control is used to define the rough relative IO cost of swapping and filesystem paging, as a value between 0 and 200.

Mic92:
Could we adapt what Fedora / PopOS are doing with zswap? I don't know what a good way would be to test this, so I would rather prefer to rely on well tested defaults from other distributions.

yajo:
Thanks for those links! It looks like zswap is indeed much better for desktop usage, mostly when I already have a swap NVMe partition.

Closing then.

https://linuxblog.io/zswap-better-than-zram/ explains more interesting details.

PR #809, opened 2026-04-28
nixos/mixins/nginx: remove dhparams

zowoq:
https://github.com/NixOS/nixpkgs/commit/95a3a84e48202a8bc624edea5726794d703a3d91

### DIFF
--- a/nixos/mixins/nginx.nix
+++ b/nixos/mixins/nginx.nix
@@ -32,18 +32,5 @@
           if config.networking.nameservers == [ ] then cloudflare else config.networking.nameservers;
       in
       map escapeIPv6 resolvers;
-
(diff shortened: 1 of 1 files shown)


Enzime (approved):
LGTM

I made the PR https://github.com/NixOS/nixpkgs/pull/515057 to drop DHE from the default cipher list according to Mozilla recommendations and remove the `services.nginx.sslDhparam` option however that PR won't get merged until after 26.05.

This should be safe to merge according to [nginx's documentation](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_dhparam)

>  By default no parameters are set, and therefore DHE ciphers will not be used.

PR #828, opened 2026-06-01
nixos/modules: add google-cloud-ops-agent

adrian-gierakowski:
This adds the Google Cloud Ops Agent and its OpenTelemetry Collector dependency, along with the NixOS module to configure and run them. The agent is used to gather logs and metrics from Google Compute Engine instances and send them to Cloud Logging and Cloud Monitoring.


zowoq:
I think this should go to nixpkgs instead, having packages and a full service module here is a bit out of scope.

Issue #337, opened 2023-12-14
openssh authorizedKeysFiles breaks nixos-anywhere


sedlund:
nixos-anywhere generates a temporary ssh key for deployment.  it puts it into $HOME/.ssh which https://github.com/nix-community/srvos/blob/cd802e2933c567ea91de48dbe8968f41a5d9a642/nixos/common/openssh.nix#L30

won't allow for security.

How can I use the nixpkgs default in my flake?

phaer:
Just to clarify: You want to re-install NixSS by running nixos-anywhere against a machine which is already running NixOS with srvos included?

> How can I use the nixpkgs default in my flake?
```
    services.openssh.authorizedKeysFiles = lib.mkOverride 40
      [ "%h/.ssh/authorized_keys" "/etc/ssh/authorized_keys.d/%u" ];
```
(`mkForce` equals `mkOverride 50`)

sedlund:
Yes thats right.  Using disko to re-layout drives / lvm / luks / filesystems.  Thanks for the tip!

Mic92:
We do not set this option in `nixos-images`, so it's unclear to me why this is causing problems with nixos-anywhere?

sedlund:
The problem comes when a system that is built with default SrvOS is initially connected to with `nixos-anywhere` cli  (to be repartitioned / reloaded etc)

SrvOS generates in `/etc/ssh/sshd_config` this line
`AuthorizedKeysFile /etc/ssh/authorized_keys.d/%u`

nixos-anyhwere will generate temp ssh keys, [ssh-copy-id](https://github.com/nix-community/nixos-anywhere/blob/daf19effbafba2cfeac4c41e17dccdd07ca86cb3/src/nixos-anywhere.sh#L318-L325)'s them to the current NixOS `$HOME/.ssh`, then attempts to [import_facts](https://github.com/nix-community/nixos-anywhere/blob/daf19effbafba2cfeac4c41e17dccdd07ca86cb3/src/nixos-anywhere.sh#L346) using the [ssh_](https://github.com/nix-community/nixos-anywhere/blob/daf19effbafba2cfeac4c41e17dccdd07ca86cb3/src/nixos-anywhere.sh#L207-L209) function which specifies these temp keys  which are not allowed.

It has been a long while since this occurred, so I believe  this is what happened.

Mic92:
Ah. Now I get it.

PR #833, opened 2026-06-09
nixos/openssh: fix ssh-copy-id and nixos-anywhere

Enzime:
This breaks reinstalling an existing machine with `nixos-anywhere` as it relies on `ssh-copy-id` injecting a temporary deployment SSH key to `~/.ssh/authorized_keys`

This rule is not enabled by default in NixOS so I think it's fine to remove

### DIFF
--- a/darwin/common/openssh.nix
+++ b/darwin/common/openssh.nix
@@ -1,12 +1,6 @@
 # Better defaults for OpenSSH
-{ lib, ... }:
 {
   environment.etc."ssh/sshd_config.d/102-srvos.conf".text = ''
-    # Only allow system-level authorized_keys to avoid injections.
-    # nix-darwin uses AuthorizedKeysCommand to set system-level keys
-    # https://github.com/LnL7/nix-darwin/blob/f61d5f2051a387a15817007220e9fb3bbead57b3/modules/programs/ssh/default.nix#L158
-    AuthorizedKeysFile none
-
     X11Forwarding no
     KbdInteractiveAuthentication no
     PasswordAuthentication no
--- a/nixos/common/openssh.nix
+++ b/nixos/common/openssh.nix
@@ -1,5 +1,4 @@
 # Better defaults for OpenSSH
-{ config, lib, ... }:
 {
   services.openssh = {
     settings.X11Forwarding = false;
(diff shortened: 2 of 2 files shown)


sedlund:
Linking #337 

To override, you have to use `mkOverride` with a lower priority than `mkForce`'s default of 50.

I think removal is probably okay.

But if it stays, maybe instead of using `mkForce`, use `mkOverride 60`, so end users can use mkForce instead of having to know that trivia, to override if needed.

zowoq:
If this is going to be removed also need to remove it on darwin:

https://github.com/nix-community/srvos/blob/bc0ca7b0f7c8ca9dc55166438fec4dc2dd74f8a4/darwin/common/openssh.nix#L5-L8

zowoq:
I'm not sure about silently removing a security setting that has been enabled since the start of this repo? I think it needs some sort of notification for users but I don't know what, changelog doesn't seem like the right method considering we've never had one and this is landing in unstable/stable at the same time.

Mic92:
I wouldn't say this breaks nixos-anywhere. But to use nixos-anywhere you have to configure your key declaratively first - you can even do this if you are building installation images. We also turn off password auth so you wouldn't be able to connect either. Yes it breaks ssh-copy-id but this by design.

sedlund:
> But to use nixos-anywhere you have to configure your key declaratively first

I originally explained the mechanism here https://github.com/nix-community/srvos/issues/337#issuecomment-2228078595 with links showing in source. 

It does break nixos-anywhere when it goes to gather facts.  It currently generates random keys then uses ssh-copy-id to copy these random keys, and then tries to connect and fails.

Enzime:
@Mic92 could you reopen this PR? the issue is not that my own key has not been declaratively pre-configured, but that since https://github.com/nix-community/nixos-anywhere/pull/522 `nixos-anywhere` will always try to use only the deployment key (to avoid potentially requiring authorization from the SSH agent again) which is copied by `ssh-copy-id` to `authorized_keys`

Mic92:
Oh boy, that's a regression in nixos-anywhere than. This used to work without problems for many years. Why does nixos-anywhere do this?

PR #836, opened 2026-06-18
roles/github-actions-runner: multi-org support

znaniye:
- **roles/github-actions-runner: support multiple orgs**
- **roles/github-actions-runner: assert on removed url/count/githubApp.login**


zowoq:
I'm proposing that we remove github-actions-runner: https://github.com/nix-community/srvos/pull/837

PR #854, opened 2026-07-21
fix: Hostname was interpreted as mathematical expression

VolodiaPG:
In the hostname change detection script, quotes were forgotten, allowing hostnames such as "home-server" to be interpreted as home minus server.


bartoostveen (approved):
Thanks, this fixes my issues while trying to name a host `delta` 😅

